PR #862, opened 2014-12-22
Monitoring server self info is not included in the count.

o-wakabayashi:
Except the self monitoring info from the count of "hosts(with problem)" and
"triggers(with problem)" in Dashboard.
(Applied the change of license.)


yoshfuji:
Thus was update of #730

PR #895, opened 2015-01-08
Use g_strerror() instead of strerror() to improve portability

ashie:
https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strerror


ashie:
Thanks!

PR #898, opened 2015-01-08
Fatal server error from FaceRest

yoshfuji:
Handle exception from FaceRest as fatal event.
Otherwise other thread continue running.

### DIFF
--- a/server/src/FaceRest.h
+++ b/server/src/FaceRest.h
@@ -82,7 +82,7 @@ protected:
 	static void handlerLogin(ResourceHandler *job);
 	static void handlerLogout(ResourceHandler *job);
 
-	int onCaughtException(const std::exception &e)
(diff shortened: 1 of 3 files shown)


server/src/FaceRest.h
```diff
@@ -82,6 +82,11 @@ class FaceRest : public FaceBase {
 	static void handlerLogin(ResourceHandler *job);
 	static void handlerLogout(ResourceHandler *job);
 
+	int onCaughtException(const std::exception &e)
```
kz0817: Please add the override keyword as

int onCaughtException(const std::exception &e) override

yoshfuji:
Ok
Since original changes did not introduce any regressions, I've pushed 2 additional commits to reflect comment rather than recreating branch.

PR #904, opened 2015-01-13
Add HistoryLoader

ashie:
Add a new JavaScript object "HistoryLoader" to make easy to load history of multiple items asynchronously.

### DIFF
--- a/client/static/js/history_view.js
+++ b/client/static/js/history_view.js
@@ -55,7 +55,7 @@ HistoryLoader.prototype.load = function() {
 
   function getHistoryQuery() {
     var query = $.extend({}, self.options.historyQuery);
-    var lastReply, lastData, now;
+    var lastReply, lastData;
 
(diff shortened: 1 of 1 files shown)


client/static/js/history_view.js
```diff
@@ -17,44 +17,276 @@
...
+
+  function shiftHistory(history) {
+    var beginTimeInMSec, data;
+
+    if (!history)
+      return;
+
+    beginTimeInMSec = (self.lastQuery.endTime - self.getTimeSpan()) * 1000;
+
+    while(history.length > 0 && history[0][0] < beginTimeInMSec) {
```
kz0817: We basically insert a space between while and '(', don't we ?

ashie: Sorry, I inserted it.

ashie:
Thank you for merging!

PR #913, opened 2015-01-15
[Issue767] Bug fix and create a test

Mnakagawa:


server/common/HatoholError.cc
```diff
@@ -97,6 +97,8 @@ void HatoholError::init(void)
 	DEFINE_ERR(USER_ROLE_NAME_OR_PRIVILEGE_FLAGS_EXIST,
 		   "The same user role name or a user role with the same "
 		   "privilege already exists.");
+	DEFINE_ERR(DELETE_MYSELF,
+		   "Can't delete login user.");
```
kz0817: A word 'login user' is ambiguous. Some people may think that it includes other users who are logged in.
How about using the following sentence.
"A user cannot delete myself."

server/test/testDBTablesUser.cc
```diff
@@ -498,6 +498,19 @@ void test_deleteUser(void)
...
 
+void test_deleteMyself(void)
+{
+	DECLARE_DBTABLES_USER(dbUser);
+	OperationPrivilege privilege(ALL_PRIVILEGES);
+	const UserIdType targetId = privilege.getUserId();
+	HatoholError err = dbUser.deleteUserInfo(targetId, privilege);
+	assertHatoholError(HTERR_DELETE_MYSELF, err);
+
+	UserIdSet userIdSet;
```
kz0817: These these lines are unnecessary, aren't they ?
The first half of the test seems not to delete a user at all.

server/test/testFaceRestUser.cc
```diff
@@ -363,17 +363,23 @@ void test_deleteUser(void)
 {
 	startFaceRest();
 
-	const UserIdType userId = findUserWith(OPPRVLG_DELETE_USER);
-	string url = StringUtils::sprintf("/user/%" FMT_USER_ID, userId);
+	const UserIdType deleterId = findUserWith(OPPRVLG_DELETE_USER);
+	const UserIdType targetId = 1;
+
+	// When deleterId and targetId are same, this test case will fail.
```
kz0817: How about using 'be identical with' words' ?

server/test/testFaceRestUser.cc
```diff
@@ -363,17 +363,23 @@ void test_deleteUser(void)
...
+	// Because, Login user can't delete myself.
```
kz0817: Basically comma ',' just after Because is unnecessary. Let's see the examples in a dictionary.
And 'login user' is ambiguous as pointed out in the above comment.

server/test/testFaceRestUser.cc
```diff
@@ -363,17 +363,23 @@ void test_deleteUser(void)
...
+	// If the test fails, you need to fix Users data of DBTableTest.cc.
```
kz0817: I feel something is wrong for the line.
If the test fails in the future, the wrong fix will have to be done.

Mnakagawa:
kz0817 san
Thank you for your pointed out.
I fixed. And please confirm.

PR #915, opened 2015-01-15
[server] ArmZabbixAPI,ConfigManager: Do not get old events at first by d...

yoshfuji:
...efault.

When adding a new zabbix server, Hatohol tried to get all events
from the server.  It may take very long time if the server has
accumulated tons of events.  This seems very harmful because it
places a high load on the monitoring server.

Do not load old events by default unless --load-old-events option
is specified.


ashie:
Thanks!

Issue #158, opened 2014-03-10
[Server] It takes time to get event


bynnchapu:
I changed the way of getting monitoring data to add hostgroup information.
But, it takes time to get event data of over 10,000 in the version of now.

Current way doesn't take time to get event data.
Because it designates number of event.

Before the change (in 1a0ca1b revision)
```
$ curl -X GET -H X-Hatohol-Session:**** "http://localhost:33194/event" | jsonpipe
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current Dload  Upload   Total   Spent    Left  Speed
100 1858k  100 1858k    0     0  6237k      0 --:--:-- --:--:-- --:--:-- 6238k
<snip>
/numberOfEvents 11556
```

After the change (in a7c0326 revision)
```
$ curl -X GET -H X-Hatohol-Session:2027e49c-9601-4db1-9520-a2b5985cfe16
"http://localhost:33194/event" | jsonpipe
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current Dload  Upload   Total   Spent    Left  Speed
100 2094k  100 2094k    0     0   146k      0  0:00:14  0:00:14 --:--:--  491k
<snip>
/numberOfEvents 11556
```

kz0817:
Currently there's no problems for the new implementation. So I close this issue.

PR #916, opened 2015-01-15
Add new index for events: TimeSec

bynnchapu:
In Hatohol DB has tens of millions of events data (occured 40 events / sec), ajax_events show events too slow.

I and @yoshfuji san investigate this problem and found cause of this issue.
As a result of investigation, the cause did not create index for events.

So, I think we need to add new index for events.

Related to #158

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -319,8 +319,14 @@ static const int columnIndexesEventsUniqId[] = {
   IDX_EVENTS_SERVER_ID, IDX_EVENTS_ID, DBAgent::IndexDef::END,
 };
 
+static const int columnIndexesEventsTimeSec[] = {
+  IDX_EVENTS_TIME_SEC, IDX_EVENTS_TIME_NS,
+  IDX_EVENTS_UNIFIED_ID, DBAgent::IndexDef::END,
+};
+
 static const DBAgent::IndexDef indexDefsEvents[] = {
   {"EventsId", (const int *)columnIndexesEventsUniqId, false},
+  {"TimeSec", (const int *)columnIndexesEventsTimeSec, false},
   {NULL}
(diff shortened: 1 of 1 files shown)


server/src/DBTablesMonitoring.cc
```diff
@@ -319,8 +319,14 @@ static const int columnIndexesEventsUniqId[] = {
...
 };
 
+static const int columnIndexesEventsTimeSec[] = {
+  IDX_EVENTS_TIME_SEC, IDX_EVENTS_TIME_NS,
+  IDX_EVENTS_UNIFIED_ID, DBAgent::IndexDef::END,
+};
+
 static const DBAgent::IndexDef indexDefsEvents[] = {
   {"EventsId", (const int *)columnIndexesEventsUniqId, false},
+  {"TimeSec", (const int *)columnIndexesEventsTimeSec, false},
```
ashie: I think adding "Sec" to the name is inappropriate because it also includes ns and unified_id.

ashie:
We need to increment the DB version & migration code.
Please call dbAgent.fixupIndexes() at updateDB() to update old DB.

server/src/DBTablesMonitoring.cc
bynnchapu: Thank you for your comment.
I'll reconsider and send Pull Request again.

ashie:
> We need to increment the DB version & migration code.
> Please call dbAgent.fixupIndexes() at updateDB() to update old DB.

It may be wrong.
Hatohol do it automatically at the constructor of DBTables?

https://github.com/project-hatohol/hatohol/blob/master/server/src/DBTables.cc#L85

bynnchapu:
I close this issue once.
I'll re-open this pull request or re-send another Pull Request when I fix (and add migration) code.

PR #932, opened 2015-01-21
[Server][ZabbixAPI] Fix a bug that add disable hosts to monitoring host

Mnakagawa:


server/common/ZabbixAPI.cc
```diff
@@ -695,6 +695,7 @@ SoupMessage *ZabbixAPI::queryHost(HatoholError &queryRet)
 	agent.startObject("params");
 	agent.add("output", "extend");
 	agent.add("selectGroups", "refer");
+	agent.add("monitored_hosts", "True");
```
bynnchapu: You should use ```JSONBuilder::addTrue()``` function.

bynnchapu: Umm...
It seems that this ```ZabbixAPI``` class does not unify to use ```add("member_name", "True")``` and ```addTrue("member_name")``` to add boolean value...

PR #940, opened 2015-01-22
[HatoholConnector] Add a missing local variable "self"

ashie:
This commit fixes issue #938.

Because the "request" property of HatoholConnector is stored to
an inappropriate object, it can't recognize the HTTP method
correctly on a certain condition.


ashie:
Thanks!

PR #941, opened 2015-01-22
[DBTablesMonitoring] Add new DB index for events: EventsTimeSequence

bynnchapu:
In Hatohol DB has tens of millions of events data (occured 40 events / sec), ajax_events show events too slow.

I and @yoshfuji san investigate this problem and found cause of this issue.
As a result of investigation, the cause did not create index for events.

So, I think we need to add new index for events.

Related to #158
Fix of #916

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -319,8 +319,14 @@ static const int columnIndexesEventsUniqId[] = {
   IDX_EVENTS_SERVER_ID, IDX_EVENTS_ID, DBAgent::IndexDef::END,
 };
 
+static const int columnIndexesEventsTimeSequence[] = {
+  IDX_EVENTS_TIME_SEC, IDX_EVENTS_TIME_NS,
+  IDX_EVENTS_UNIFIED_ID, DBAgent::IndexDef::END,
+};
+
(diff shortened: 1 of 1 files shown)


ashie:
> As a result of investigation, the cause did not create index for events.

Supplement:
The cause is "ORDER BY " clause.
Although the SQL statement has "LIMIT" clause, "ORDER BY" without index causes searching all rows.

bynnchapu:
> Supplement:
> The cause is "ORDER BY " clause.
> Although the SQL statement has "LIMIT" clause, "ORDER BY" without index causes searching all rows.

Thank you for your supplement !

ashie:
Memo:
Indexes are always recreated (if it's needed) on relaunching Hatohol server.

bynnchapu:
> Memo:
> Indexes are always recreated (if it's needed) on relaunching Hatohol server.

Yes.
@kz0817 san said that.

bynnchapu:
Thanks!

kz0817:
> Memo:
> Indexes are always recreated (if it's needed) on relaunching Hatohol server.
> Yes.
> @kz0817 san said that.

'recreate' is a misleading word.
Precisely speaking, indexes are created if they don't exist and
existing ones are deleted if they're no longer used.

PR #946, opened 2015-01-23
[Server][DBTableMonitoring] Add a update processing of hostname.

Mnakagawa:


server/src/DBTablesMonitoring.cc
```diff
@@ -2101,8 +2099,16 @@ void DBTablesMonitoring::updateHosts(const HostInfoList &hostInfoList,
...
+		if (currHostItr != currValidHosts.end()) {
+			const HostInfo &existHost = *currHostItr->second;
+			if (existHost.hostName != newHostInfo.hostName){
+				updatedHostInfoList.push_back(newHostInfo);
+				continue;
+			}
+		}
 		if (currValidHosts.erase(newHostInfo.id) >= 1) {
-			// The host already exits. We have nrothing to do.
+			// The host already exits. We have to do nothing.
```
kz0817: Please fix the comment in another commit.

server/src/DBTablesMonitoring.cc
```diff
@@ -2101,8 +2099,16 @@ void DBTablesMonitoring::updateHosts(const HostInfoList &hostInfoList,
...
 		const HostInfo &newHostInfo = *newHostsItr;
+		HostIdHostInfoMapConstIterator currHostItr = currValidHosts.find(newHostInfo.id);
+		if (currHostItr != currValidHosts.end()) {
+			const HostInfo &existHost = *currHostItr->second;
+			if (existHost.hostName != newHostInfo.hostName){
+				updatedHostInfoList.push_back(newHostInfo);
+				continue;
+			}
+		}
 		if (currValidHosts.erase(newHostInfo.id) >= 1) {
```
kz0817: This code does the search twice (find() and erase()) when an element in currValidHosts is removed.
erase() accepts an iterator that is directly pointing an element.

I.e.
currHostItr should be passed to erase() when it should be removed.

kz0817:
Please add a unit test to check the feature.

Mnakagawa:
I create a new pull request.
Please, see that.

PR #947, opened 2015-01-23
[ZabbixAPI] Use addTrue() function instead of add() function.

bynnchapu:
I found a mistake in #934.

We should use ```addTrue()``` or ```addFalse()``` function when we set value of JSON as boolean value.


yoshfuji:
Hmm, why do we have special methods, addTrue() and addFalse()?
I'd say we should use add(), instead of such special methods.

ashie:
> Hmm, why do we have special methods, addTrue() and addFalse()?
> I'd say we should use add(), instead of such special methods.

Because calling add(string, int) is ambiguous (maybe).

PR #951, opened 2015-01-23
[TriggersView] Fix an inappropriate absolute path

ashie:
Fix issue #944


ashie:
I merged it myself because it's a very easy fix.

PR #956, opened 2015-01-28
 Fix a script as not to display a group that doesn't include monitoring host.

Mnakagawa:


client/static/js/dashboard_view.js
```diff
@@ -181,6 +188,8 @@ var DashboardView = function(userProfile) {
           html += escapeHTML(replyData["serverStatus"][x]["serverNickname"]);
           html += "</td>";
         }
+        if(hostStatus["numberOfGoodHosts"] + hostStatus["numberOfBadHosts"] == 0)
```
ashie: Please insert a space between "if" and "(".

client/static/js/dashboard_view.js
```diff
@@ -134,18 +134,25 @@ var DashboardView = function(userProfile) {
...
+        hostStatus = hostStatuses[y];
         html += "<tr>"; // ==============  start of a row ================
         if (y == 0) {
           html += "<td rowspan='" + escapeHTML(parsedData[serverId]["numberOfHostgroups"]) + "'>";
           html += escapeHTML(replyData["serverStatus"][x]["serverNickname"]);
           html += "</td>";
         }
+        if (hostStatus["numberOfGoodHosts"] + hostStatus["numberOfBadHosts"] == 0) {
+          ++y;
+          continue;
```
ashie: It seems that it breaks the "tr" element.
Although the <tr> tag is opened at the top of the block, it will never be closed when the "continue" is evaluated.
(It's closed at the end of the block).

client/static/js/dashboard_view.js
```diff
@@ -181,6 +188,8 @@ var DashboardView = function(userProfile) {
           html += escapeHTML(replyData["serverStatus"][x]["serverNickname"]);
           html += "</td>";
         }
+        if(hostStatus["numberOfGoodHosts"] + hostStatus["numberOfBadHosts"] == 0)
+          continue;
```
ashie: It also breaks the "tr" element.

Mnakagawa:
@ashie san
I fixed their.
How does the look? I have no confidence.

Mnakagawa:
I create a new pull request.
Close it.

PR #965, opened 2015-01-29
[client] hatohol_pager: Add missing function-local variable self.

yoshfuji:
This fixes following Webpage Error with Internet Explorer:
| Error: Object doesn't support property or method 'update'


ashie:
Thanks!

PR #973, opened 2015-02-02
When Qpid daemon is absent, continue to update the WebUI.

o-wakabayashi:
Add exception handling of Send (Exception occurs when Qpidd is stopped).
Fixed processing of response received.

### DIFF
--- a/server/common/HatoholArmPluginInterface.cc
+++ b/server/common/HatoholArmPluginInterface.cc
@@ -157,6 +157,10 @@ struct HatoholArmPluginInterface::Impl {
 		connected = false;
 		hapi->onSessionChanged(NULL);
 	}
+	
+	bool connectstat(void) {
+		return connected;
+	}
 
 	static void destroyReplyWaiter(ReplyWaiter *replyWaiter,
 	                               Impl *impl)
@@ -290,6 +294,11 @@ void HatoholArmPluginInterface::send(
 	request.setReplyTo(m_impl->receiverAddr);
 	request.setContent(smbuf.getPointer<char>(0), smbuf.size());
 	m_impl->sender.send(request);
+}
+
+bool HatoholArmPluginInterface::getConnectStat(void)
+{
+	return m_impl->connectstat();
 }
 
 bool HatoholArmPluginInterface::getMessagingContext(MessagingContext &msgCtx)
--- a/server/common/HatoholArmPluginInterface.cc
+++ b/server/common/HatoholArmPluginInterface.cc
@@ -158,7 +158,8 @@ struct HatoholArmPluginInterface::Impl {
 		hapi->onSessionChanged(NULL);
 	}
 	
-	bool connectstat(void) {
+	bool getconnectedstatus(void) {
+		AutoMutex autoMutex(&connectionLock);
 		return connected;
 	}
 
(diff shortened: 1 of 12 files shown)


server/common/HatoholArmPluginInterface.cc
```diff
@@ -157,6 +157,10 @@ struct HatoholArmPluginInterface::Impl {
 		connected = false;
 		hapi->onSessionChanged(NULL);
 	}
+	
+	bool connectstat(void) {
```
kz0817: The name is unreadable.
Some people may think it makes a connection because it begin with a verb: connect.

server/common/HatoholArmPluginInterface.cc
```diff
@@ -157,6 +157,10 @@ struct HatoholArmPluginInterface::Impl {
...
+		return connected;
```
kz0817: The member: connected has to be accessed with the lock: connectionLock in order to avoid races.

server/src/DataStore.cc
```diff
@@ -30,8 +30,9 @@ void DataStore::setCopyOnDemandEnable(bool enable)
 {
 }
 
-void DataStore::startOnDemandFetchItem(Closure0 *closure)
+bool DataStore::startOnDemandFetchItem(Closure0 *closure)
 {
+	return true;
```
ashie: I think it should return the result of isFetchItemsSupported() by default.

server/common/HatoholArmPluginInterface.cc
o-wakabayashi: Yes, I understand.
"connected" will be changed from other threads.
Therefore, lock(connectionLock) is required.

server/common/HatoholArmPluginInterface.cc
```diff
@@ -157,6 +157,11 @@ struct HatoholArmPluginInterface::Impl {
 		connected = false;
 		hapi->onSessionChanged(NULL);
 	}
+	
+	bool getconnectedstatus(void) {
```
cosmo0920: Hmm..., Hatohol project uses camelCase naming rule.
Please use `getConnectedStatus` instead of `getconnectedstatus`.

server/src/DataStore.cc
o-wakabayashi: Yes, I understand...
Usually, override function is executed.

server/common/HatoholArmPluginInterface.cc
kz0817: I propose 'isConnected()'.
'getConnectedStatus()' make people think of getting a connected status.
And, status can be any values (2 or more) in general. However, the return type of the method is bool.

o-wakabayashi: Thank you for comments. I use the isConnected().
Because, return type of the method is bool.
(Function name is used many "is ~".)

server/common/HatoholArmPluginInterface.cc
```diff
@@ -292,6 +297,11 @@ void HatoholArmPluginInterface::send(
 	m_impl->sender.send(request);
 }
 
+bool HatoholArmPluginInterface::getConnectStat(void)
```
ashie: How about also rename it to "isConnected()"?

o-wakabayashi: Thank you for comments and sorry for late.
I will similarly modified(it is easy to understand.).

ashie:
Confirmed:

* Can build on Ubuntu-14.04 & Scientific Linux 6.3.
* Show "Overview items" & "Latest view" immediately with old data when the qpid daemon is absent.
* Data on these pages are updated when the qpid daemon revives.

ashie:
Thanks!

o-wakabayashi:
Thank you for merged and many comment.

PR #977, opened 2015-02-02
Clean up history view

ashie:
* Remove redundant properties of HistoryView ("endTime" and "timeSpan") because the "timeRange" object has same information
* Replace the word "refresh" with "reload" to unify them
* Unify calling $.plot.dateGenerator()


ashie:
Thank you for merging!

PR #978, opened 2015-02-02
[test][client] Replace wrong arguments for child classes of HatoholMonit...

ashie:
...oringView

Although the first argument of child classes of HatoholMonitoring
view is userProfile, DOMElement was wrongly used.


cosmo0920:
:+1:

PR #980, opened 2015-02-03
[DBAgent] Fix the problem in which updating an index fails.

kz0817:
When a combination of columns for an index is changed,
DBAgent tries to create the new one. However, it fals due
to existence of the previous one.

This patch removed the existing index with the same name as
the one that will be created before the creation.

### DIFF
--- a/server/src/DBAgent.cc
+++ b/server/src/DBAgent.cc
@@ -350,34 +350,49 @@ void DBAgent::dropTable(const std::string &tableName)
 
 void DBAgent::fixupIndexes(const TableProfile &tableProfile)
 {
-	typedef map<string, IndexInfo *>  IndexSqlInfoMap;
-	typedef IndexSqlInfoMap::iterator IndexSqlInfoMapIterator;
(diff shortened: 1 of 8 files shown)


ashie:
Confirmed:

* Succeeded to build on Ubuntu-14.04 & Scientific Linux 6.3
* The test test_fixupSameNameIndexes() fails without this fix:
```
Error: testDBAgentMySQL::test_fixupSameNameIndexes
Unhandled C++ standard exception is thrown: <16HatoholException>: <DBAgentMySQL.cc:625> Failed to query: CREATE INDEX testIndex ON test_table (name,time): (1061) Duplicate key name 'testInde
```
```
===============================================================================
Error: testDBAgentSQLite3::test_fixupSameNameIndexes
Unhandled C++ standard exception is thrown: <16HatoholException>: <DBAgentSQLite3.cc:329> Failed to exec: 1, index i_test_table_testIndex already exists, CREATE INDEX i_test_table_testIndex ON test_table(name,time)

===============================================================================
```
* These tests succeeded with this fix.
* It works fine with old Htaohol DB.

PR #981, opened 2015-02-04
Use new host management rebase 2

kz0817:
This pull request replaces host management from DBTablesMonitoring to DBTablesHost,
which is the next generation host management sub system. DBTablesHost handles
Host ID and Hostgroup ID as string in order to store the UUID type ID such as OpenStack.

Hereafter I will replace some IDs such as Trigger ID and Event ID for string.


ashie:
:scream:

cosmo0920:
:scream_cat:

ashie:
Could you solve the following build problem?

```
make[3]: Entering directory `/home/aho/Projects/Miracle/hatohol/server/src'
  CXX      HatoholArmPluginGateJSON.lo
HatoholArmPluginGateJSON.cc: In member function 'void AMQPJSONMessageHandler::initializeHosts()':
HatoholArmPluginGateJSON.cc:86:16: error: 'class DBTablesMonitoring' has no member named 'getHostInfoList'
   dbMonitoring.getHostInfoList(hostInfoList, option);
                ^
HatoholArmPluginGateJSON.cc: In member function 'HostIdType AMQPJSONMessageHandler::findOrCreateHostID(const string&)':
HatoholArmPluginGateJSON.cc:149:16: error: 'class DBTablesMonitoring' has no member named 'addHostInfo'
   dbMonitoring.addHostInfo(&hostInfo);
                ^
In file included from DataQueryContext.h:26:0,
                 from DataQueryOption.h:27,
                 from DBTablesMonitoring.h:25,
                 from ThreadLocalDBCache.h:23,
                 from HatoholArmPluginGateJSON.cc:25:
OperationPrivilege.h: At global scope:
OperationPrivilege.h:106:37: warning: 'ALL_PRIVILEGES' defined but not used [-Wunused-variable]
 const static OperationPrivilegeFlag ALL_PRIVILEGES = 
```

server/src/DBTablesUser.cc
```diff
@@ -128,8 +128,8 @@ static const ColumnDef COLUMN_DEF_ACCESS_LIST[] = {
 	NULL,                              // defaultValue
 }, {
 	"host_group_id",                   // columnName
-	SQL_COLUMN_TYPE_BIGUINT,           // type
-	20,                                // columnLength
+	SQL_COLUMN_TYPE_VARCHAR,           // type
```
ashie: We don't support auto migration for it?

kz0817:
Oh!. I checked without HAVE_RABBITMQ. So the source file hasn't been built.
I'll work on it and let you know after it's completed.

server/src/DBTablesUser.cc
kz0817: In fact, I'm still thinking.
I have columns of which I'd like to change the type other than here, such as Trigegr ID, and Event ID.
Of course, providing automatic upgrade feature is better. However, it takes a time.
I don't known what is important for us. Adding features rapidly, Compatibility, Easy to use ?
If we add an automatic migration feature, we can do it after. So shall we ignore that in this pull request.

kz0817:
I fixed the build error on HatoholArmPluginGateJSON.cc
Please review the request again.

server/src/DBTablesUser.cc
ashie: I think we don't always have to support auto migration.

But when we breaks compatibility, we have to notice it users & other developers.
This modification works with old Hatohol DB without warnings and almost works fine.
But it silently contains an incompatibility issue concerned with this column.
It's not good.

How about reject upgrading, show the reason and stop the program at updateDB()?

kz0817:
I close this issue once. According to the discussion in the yesterday's face2face meeting, after I merge the feature that prevents Hatohol server from booting if any of the major DBTalbes version changes, I will send a pull request again.

Issue #979, opened 2015-02-03
[client] Add expires and no-cache in the response


kz0817:
YOSHIFUJI-san said that 'Log out' doesn't work when IE (operator side) and apache are used.
And he thinks some caching mechanism causes the problem and asked me how to add expires and no-cache in an HTML header.

kz0817:
I think
For HTML files, they should be inserted into client/viewer/base_ajax.html
For APIs to Hatohol server, we should fix client/hatohol/forwardview.py

kz0817:
The header elements on the Django's framework seems to be inserted as below according to http://www.djangoproject.jp/doc/ja/1.0/ref/request-response.html#django.http.HttpResponse
>>> response = HttpResponse()
>>> response['Pragma'] = 'no-cache'

yoshfuji:
Thank you.  So far, I use mod_expires apache module as follows:
ExpireDefault "now"
ExpireActive on

cosmo0920:
For now, #994 and #1001 are merged into master.
This issue is fixed?
Or, there is an issue?

PR #994, opened 2015-02-06
[client] Add no-cache headers in response

cosmo0920:
* Django 1.5: https://docs.djangoproject.com/en/1.5/ref/request-response/#setting-headers

#979


kz0817:
Is there any reason that "Pragma" and "Expires" are not set in python code ?

cosmo0920:
It is no reason that "Pragma" and "Expires" are not to set in python code.
It is just forgot it.

kz0817:
I confirmed that the added ones were in the header. 
tunnel

cosmo0920:
Thanks!

PR #996, opened 2015-02-10
[testHapProcessStandard] Fix a race that sometimes makes tests fail.

kz0817:
A GLibMain loop instance is wrongly used by two threads. So the
unexpected behavior has happened.


ashie:
Memo:
* m_localLoop: The main loop for the caller of HatoholArmPluginGate::startOnDemandFetchItem() (it's a FaceRest thread in the real Hatohol server)
* loop: The main loop for the HAP

PR #1001, opened 2015-02-16
Expires header must be an HTTP_date, not 0.

yoshfuji:


cosmo0920:
Thanks. I've confirmed RFC1123 dose not allow number for `Expires` header.

Section 14.21 Expires says:

> The format is an absolute date and time as defined by HTTP-date in
   section 3.3.1; it MUST be in RFC 1123 date format:

But also says following exception clause:

>  HTTP/1.1 clients and caches MUST treat other invalid date formats,
   especially including the value "0", as in the past (i.e., "already
   expired").

Value "0" means `already expires` in HTTP/1.1.
So, in HTTP/1.1 Expires = "0" within HTTP header treats valid value in HTTP client.

yoshfuji:
Server must not send such invalid values.  That's all.

cosmo0920:
Yes, server must not send such invalid values.
But hatohol server reply with HTTP/1.0 with Django test server.
I've also checked RFC1945(HTTP/1.0).

HTTP/1.0([RFC 1945](http://tools.ietf.org/html/rfc1945)) says:

[Note in section 10.7](http://tools.ietf.org/html/rfc1945#section-10.7)

>  Note: Applications are encouraged to be tolerant of bad or
      misinformed implementations of the Expires header. A value of zero
      (0) or an invalid date format should be considered equivalent to
      an "expires immediately."

Hmm, this note is application side(client side).
Even if `Expires = "0"` is invalid in HTTP/1.0.

PR #1005, opened 2015-02-17
[TriggersView] Don't colorize non-fired triggers

ashie:
Because users may misunderstood it as problem triggers.


cosmo0920:
It seems to be looks good. :star2: 

I've confirmed following:

* don't colorize non-fired triggers in triggers page.

But I found another remaining problem:

* colorize non-fired event in events page.

ashie:
> But I found another remaining problem:
> 
>     colorize non-fired event in events page.

Sorry, I forgot it.
I'll modify it too.

PR #1006, opened 2015-02-17
[EventsView] Don't colorize severity labels of good events

ashie:
Because users may misunderstand them as bad events.


cosmo0920:
It seems to be looks good. :clap: 

> Because users may misunderstand them as bad events.

Yes, I'm one of them.
I've surprised about it.

Issue #1009, opened 2015-02-18
[request]Format of the HAPI-JSON


satomi-0929:
When I register an event in HAPI-JSON, I want to appoint the status.

I want you to add the variable of "the status" to the following items to notify manager of the event (ex.OK, unknown) except the problem.

https://github.com/project-hatohol/hatohol/wiki/HAPI-JSON

event
event type message must has the following format:

{
  "type": "event",
  "body": {
    "id": ${Event ID as an integer},
    "timestamp": ${Event occurred time as UNIX time in float or ISO 8601 format string},
    "hostName": "${Host name where event is occurred as string}",
    "content": "${Event detail as string}",
    "severity": "${Severity of the event (optional) (default: unknown)}"
  }
}

kz0817:
Thank you for the suggestion.
You mean you want to add an event that is not a abnormal thing (recovery or just information), don't you ?

satomi-0929:
Yes,I do.

kou:
It's reasonable.

What about the following API?

```text
{
  "type": "event",
  "body": {
    ...,
    "type": ${"good" OR "bad" OR "unknown",
    ...
  }
}
```

If you specifies `"type": "bad"`, it's the current behavior.

satomi-0929:
Thank you for your comment.
I try it.

kou:
Ah, it's not implemented yet.
It's just new API idea.

kou:
I implemented the feature at #1011.
(It's not merged yet.)

kou:
It's merged.
You can try it on master.

Thanks for your request!

satomi-0929:
I'm sorry for the late response.
Thank you for correspondence！　:-)

PR #1011, opened 2015-02-18
HAPI-JSON supports event type

kou:
It implements a request at #1009.

New HAPI JSON format:

```text
{
  "type": "event",
  "body": {
    "id": ${Event ID as an integer},
    "timestamp": ${Event occurred time as UNIX time in float or ISO 8601 format string},
    "hostName": "${Host name where event is occurred as string}",
    "content": "${Event detail as string}",
    "severity": "${Severity of the event (optional) (default: unknown)}",
    "type": "${Type of the event (optional) (default: bad)}"
  }
}
```

Optional `$.body.type` parameter is added. Available values are `good`, `bad` (default) and `unknown`. They are mapped to `EVENT_TYPE_GOOD`, `EVENT_TYPE_BAD` and `EVENT_TYPE_UNKNOWN`.

### DIFF
--- a/server/src/GateJSONEventMessage.cc
+++ b/server/src/GateJSONEventMessage.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 Project Hatohol
+ * Copyright (C) 2014-2015 Project Hatohol
  *
  * This file is part of Hatohol.
  *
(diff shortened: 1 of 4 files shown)


kz0817:
Although the test on TravisCI failed, it passed on my local computer.

kou:
> Although the test on TravisCI failed, it passed on my local computer.

Some tests (or implementations) may have memory related bugs...

PR #1013, opened 2015-02-19
The auto refresh feature is enabling when press "To next" button.

Mnakagawa:
And I delete garbage code that was not deleted by me when pull request #1003 .


ashie:
Confirmed:

* The auto realod button becomes white & the auto reload feature is certainly disabled when I click the page 2 or later.
* It becomes blue & the feature is certainly enabled when I click the page 1 or "Latest events" button.
* The auto reload feature is certainly enabled/disabled when I click the auto reload button.

Mnakagawa:
Thank you!

PR #1017, opened 2015-02-23
Db major version

kz0817:
These patches add a mechanism to check DBTables's major versions.
And if one of the major versions changes, it throws an exception. As a
result, hatohol server stop at the boot.


ashie:
Confirmed:

* It doesn't break the current behaviour.
* The new versioning mechanism seems work fine in the test code.

PR #1018, opened 2015-02-23
[Travis] Reduce needless build

ashie:
* Build at `script` instead of `before_script` to reduce needless build
* Exclude some needless test patterns
* Start slower tests ASAP
* Don't use the make option `-j` for distcheck to try to avoid compiler error


cosmo0920:
:+1:

PR #1019, opened 2015-02-23
Use new host management rebase 4

kz0817:
In #981, @ashie said 'We don't support auto migration for it?'
And then we discussed the issue. So we've introduced the new concept a major version of DBTables.
In this pull request, the major version of DBTablesUser in which a type type is changed bumped up.


ashie:
The hostgroup filter for items doesn't work because it still uses tableProfileMapHostsHostgroups.

I confirmed that It works fine with the following patch:

```diff
diff --git a/server/src/DBTablesMonitoring.cc b/server/src/DBTablesMonitoring.cc
index 38317cf..db67cc8 100644
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -1311,7 +1311,7 @@ static const HostResourceQueryOption::Synapse synapseItemsQueryOption(
   tableProfileItems, IDX_ITEMS_ID, IDX_ITEMS_SERVER_ID,
   tableProfileItems, IDX_ITEMS_HOST_ID,
   true,
-  tableProfileMapHostsHostgroups,
+  tableProfileHostgroupMember,
   IDX_MAP_HOSTS_HOSTGROUPS_SERVER_ID, IDX_MAP_HOSTS_HOSTGROUPS_HOST_ID,
   IDX_MAP_HOSTS_HOSTGROUPS_GROUP_ID);
 ```

**Comment by ashie** [comment]
> I confirmed that It works fine with the following patch:

If there is no problem, please commit it.

**Comment by kz0817** [comment]
>> I confirmed that It works fine with the following patch:
> If there is no problem, please commit it.

Thank you for the patch. I committed it to the branch.
Please resume the review.

File: server/src/HatoholArmPluginGateJSON.cc [hunk 1]
```diff
@@ -119,7 +115,7 @@ class AMQPJSONMessageHandler : public AMQPMessageHandler
 		eventInfo.serverId = m_serverInfo.id;
 		eventInfo.id = message.getID();
 		eventInfo.time = message.getTimestamp();
-		eventInfo.type = message.getType();
+		eventInfo.type = EVENT_TYPE_BAD;
```
ashie: Probably it's degraded.

kou: Yes. The change should not be included.

server/src/HostResourceQueryOption.cc
```diff
@@ -161,9 +160,9 @@ string HostResourceQueryOption::getCondition(void) const
 		if (m_impl->targetHostgroupId != ALL_HOST_GROUPS) {
 			addCondition(condition,
 			  StringUtils::sprintf(
-				"%s=%s",
+				"%s='%" FMT_HOST_GROUP_ID "'",
 				getHostgroupIdColumnName().c_str(),
-				dbTermCodec->enc(m_impl->targetHostgroupId).c_str())
+				m_impl->targetHostgroupId.c_str())
```
ashie: It has SQL Injection vulnerabilities.

kz0817:
> It has SQL Injection vulnerabilities.
:fearful:
Oh!! Thank you for the point out.

server/src/DBTablesUser.cc
```diff
@@ -753,13 +757,14 @@ HatoholError DBTablesUser::addAccessInfo(AccessInfo &accessInfo,
...
-	  "%s=%" FMT_USER_ID " AND %s=%" PRIu32 " AND %s=%" PRIu64,
+	  "%s=%" FMT_USER_ID " AND %s=%" PRIu32 " AND "
+	  "%s='%" FMT_HOST_GROUP_ID "'",
 	  COLUMN_DEF_ACCESS_LIST[IDX_ACCESS_LIST_USER_ID].columnName,
 	  accessInfo.userId,
 	  COLUMN_DEF_ACCESS_LIST[IDX_ACCESS_LIST_SERVER_ID].columnName,
 	  accessInfo.serverId,
 	  COLUMN_DEF_ACCESS_LIST[IDX_ACCESS_LIST_HOST_GROUP_ID].columnName,
-	  accessInfo.hostgroupId);
+	  accessInfo.hostgroupId.c_str());
```
ashie: All similar codes should escape quotes in hostgroupId string.

kz0817:
I'll implement the feature in which dbTermCodec->enc() does the following. How do you think of it ?
- check if quotation is included. If so, enc() quote it.

ashie:
> I'll implement the feature in which dbTermCodec->enc() does the following. How do you think of it ?

Sounds good!

kz0817:
I fixed the problem of an SQL injection in #1036. So I close this pull request.

PR #1022, opened 2015-02-24
[Graph] Add UI for selecting items

ashie:
You can choose items to plot onto a graph by a dialog.
The dialog can be open by pressing the gear button on a history page.
It will be opened automatically if the page has no item
(open http://url.of.hatohol/ajax_history without query parameters).

Unfortunately there is no way to save the configuration yet.
And also there is no way to customize other settings (e.g. axes parameters).
I'm now working on them.

customize-button

customization-ui

### DIFF
--- a/client/static/js/hatohol_monitoring_view.js
+++ b/client/static/js/hatohol_monitoring_view.js
@@ -19,6 +19,7 @@
 
 window.onerror = function(errorMsg, fileName, lineNumber) {
   var place = "[" + fileName + ":" + lineNumber + "]";
+  console.log(place + " " + errorMsg);
   HatoholMonitoringView.prototype.setStatus({
     "class": "Danger",
     "label": gettext("Error"),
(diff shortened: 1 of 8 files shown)


client/static/js/hatohol_monitoring_view.js
```diff
@@ -19,6 +19,7 @@
 
 window.onerror = function(errorMsg, fileName, lineNumber) {
   var place = "[" + fileName + ":" + lineNumber + "]";
+  console.error(place + " " + errorMsg);
```
kz0817: Is this really necessary ?

ashie: I noticed that sometimes trailing async functions overwrite the previous status and discard 
the previous error messages immediately (when a trailing async function doesn't have errors, 
the status becomes "DONE" and the button becomes green). It's hard to debug. 

So I want to output all error messages to console.

kz0817: I see. Thank you for the explanation
BTW, will it make IE support difficult ?

ashie: It works with IE11 on my environment.
But I don't test it with IE10 or older yet.

kz0817: OK. Now we don't have a standard way to check it. So let's put the issue off.

ashie:
Thank you for merging!

PR #1024, opened 2015-02-24
Dbterm codec ver up

kz0817:
These patches adds the following feature.
- Create a string value for LHS of an SQL statement, in which quotations are escaped to prevent an SQL injection.
- Add a utilities class that safely provide C-style string for printf() families. This helps code to create an SQL statement simple.

### DIFF
--- a/server/src/DBTermCStringProvider.cc
+++ b/server/src/DBTermCStringProvider.cc
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2015 Project Hatohol
+ *
+ * This file is part of Hatohol.
+ *
+ * Hatohol is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License, version 3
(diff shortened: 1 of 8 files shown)


kz0817:
> I'm wondering why this class stores previous strings.
> Could you tell me the reason?

If a string object isn't stored, it's destroyed at the end of the function
in some situations and the pointer obtained by c_str() points the invalid region.

For example, the following 'ptr' points an invalid region.

const char *getChar(const string &s)
{
    return s.c_str();
}

DBTermCodec codec;
const char *ptr = getChar(codec.enc("hoge"));
printf("%s", getChar(codec.enc("foo")); // This also passes an invalid (dangling) pointer.

kz0817:
I come up with the more general idea.
I once close the pull request.

kz0817:
> // This also gets a dangling pointer.
> printf("%s", getChar(codec.enc("foo"));

Is this OK?

PR #1026, opened 2015-02-25
[DBTermCStringProvider] Add a new class.

kz0817:
This class provides a method that returns a C-style string (const char *).
This is covenient to be used in conjunction with printf families
to make an SQL statement.

 For example,

[Previously]
StringUtils::sprintf("column=%s", dbTermCodec.enc(value).c_str());

[New]
StringUtils::sprintf("column=%s", dbTermProvider(value));

### DIFF
--- a/server/src/DBTermCStringProvider.cc
+++ b/server/src/DBTermCStringProvider.cc
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2015 Project Hatohol
+ *
+ * This file is part of Hatohol.
+ *
+ * Hatohol is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License, version 3
+ * as published by the Free Software Foundation.
+ *
+ * Hatohol is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


kz0817:
After all, it results in the same as that in the pull request #1024.

ashie:
In #1024:

> If a string object isn't stored, it's destroyed at the end of the function
> in some conditions and the pointer obtained by c_str() points the invalid region.

My question is "Why it uses list<string> instead of string?".

kz0817:
>>    If a string object isn't stored, it's destroyed at the end of the function
>>    in some conditions and the pointer obtained by c_str() points the invalid region.
>
>My question is "Why does it use list<string> instead of string as a member of this class?".

I assume the following usage.
<pre>
DBTermCStringProvier termProvider(dbTermCodec);
printf("a=%s AND b= %s", termProvider(hostId), termProvider(hostGroupId));
</pre>
In this case, the object has to hold multiple strings.

ashie:
> In this case, the object has to hold multiple strings.

I got it.

PR #1028, opened 2015-02-25
Use dbterm codec

kz0817:
Generating a part of an SQL string depends on the DB being used.
So direct reference of the DBTermCodec instance defind statically
is not undesirable.


kz0817:
I forgot to attach a file. I close this request once.

PR #1033, opened 2015-02-25
Use DBTermCStringProvider.

kz0817:
DBTermCStringProvider is a wrapper for DBTermCodec::enc().
Using this, the code becomes simple.

It also removes quotations of value of RHS in SQL statements.
As a result, we can write an condition of an SQL statement
independent of the type of the RHS value.

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -27,6 +27,7 @@
 #include "Params.h"
 #include "ItemGroupStream.h"
 #include "DBClientJoinBuilder.h"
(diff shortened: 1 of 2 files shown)


kz0817:
I close it once.

I'll replace the instance name: 'dbTerm' for 'rhs'
(it means Right Hand Side of a condition equation
in an SQL statement) in order too avoid readers from
being confused.
In this project,instances of DBTables and the child
classes have been named dbXXXX such as dbMonitoring
and dbHost.

'rhs' seems to be a too short. However, lines to generate
an SQL statement tend to be a long. So I believe the short
name makes the code simple.

PR #1034, opened 2015-02-26
Use dbterm cstr provider 2

kz0817:
This patches mainly use DBTermCStringProvider that is a helper
class of DBTermCodec. Its instance is named 'rhs' that means
Right Hand Side of a condition equation in an SQL statement.
The name also aims to avoid readers from being confused.
In this project, instances of DBTables
and the child classes have been named dbXXXX such as
dbMonitoring and dbHost.

The name 'rhs' seems to be a too short and meaningless.  However,
lines to generate an SQL statement tend to be a long. So I believe
the short name makes the code simple after all.

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -27,6 +27,7 @@
 #include "Params.h"
 #include "ItemGroupStream.h"
 #include "DBClientJoinBuilder.h"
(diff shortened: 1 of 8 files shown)


kz0817:
The fix for Makefile.am is missing. I close this pull request and send it again after the fix.

PR #1036, opened 2015-02-26
Use new host management rebase 7

kz0817:
This is a revised version of #1019, in which @ashie san pointed out the problem of an SQL injection.
They were fixed in this pull request.


ashie:
Done!（どーん！）

cosmo0920:
:boom: Boom!!!

ashie:
I confirmed following things:

* Can't launch Hatohol server with an old DB (it's the expected result):
```
[CRIT] <main.cc:175> Failed to create a database object. This program is aborted. Reason: <DBTables.cc:293> Invalid Major version: DB/Prog: 00.04 [00]/01.00 [00] (tablesId: 48)
```
* Can't inject SQL with a following hostgroup ID:
```
hostgroupId = "'; drop database hatohol; ...";
```
* All tests are passed.
* It seems that all pages works fine.

kz0817:
iyh

PR #1039, opened 2015-02-27
Hostid to str5

kz0817:
This pull request replaces HostID from integer to string to accept Host IDs
that are handled as a UUID.


kz0817:
I close this once.
I'll divide this request into some requests.

PR #1041, opened 2015-02-27
Mlpl smart buffer ver up 2

kz0817:
These patches add features to insert string into an SmartBuffer object.


ashie:
Tests passed on my environment.

PR #942, opened 2015-01-22
Exclude items of invalid hosts.

o-wakabayashi:
Exclude Items of invalid hosts ,In the same way as Trigger.
From "Overview : Items" and "latest data".

Dealing with isseue#20(https://github.com/project-hatohol/hatohol/issues/20) comments.

PR #1043, opened 2015-02-27
[DBTablesMonitoring] Fix a problem in which pseudo hosts.

kz0817:
This just filter removed hosts and passes pseudo hosts
such as self monitoring hosts.


server/src/DBTablesMonitoring.cc
```diff
@@ -1235,10 +1235,10 @@ string TriggersQueryOption::getCondition(void) const
 		addCondition(
```
ashie: Memo for me:
This code is introduced at #942.

PR #1045, opened 2015-03-02
Arm utils 2

kz0817:
These patches extract similar lines in ArmBase and HatoholArmPluginGate to a new class ArmUtils.

### DIFF
--- a/server/src/ArmUtils.cc
+++ b/server/src/ArmUtils.cc
@@ -46,3 +46,24 @@ void ArmUtils::createTrigger(
 	triggerInfoList.push_back(triggerInfo);
 }
 
+void ArmUtils::createEvent(
+  const MonitoringServerInfo &svInfo,
+  const ArmTrigger &armTrigger, EventInfoList &eventInfoList)
+{
+	EventInfo eventInfo;
+
+	eventInfo.serverId = svInfo.id;
+	eventInfo.id = DISCONNECT_SERVER_EVENT_ID;
+	eventInfo.time = SmartTime(SmartTime::INIT_CURR_TIME).getAsTimespec();
+	eventInfo.hostId = MONITORING_SERVER_SELF_ID;
+	eventInfo.triggerId = armTrigger.triggerId;
+	eventInfo.severity = TRIGGER_SEVERITY_EMERGENCY;
+	eventInfo.status = armTrigger.status;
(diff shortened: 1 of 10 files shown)


server/src/HatoholArmPluginGate.cc
```diff
@@ -517,70 +457,15 @@ void HatoholArmPluginGate::setPluginAvailabelTrigger(const HatoholArmPluginWatch
 void HatoholArmPluginGate::setPluginConnectStatus(const HatoholArmPluginWatchType &type,
 					     const HatoholArmPluginErrorCode &errorCode)
 {
-	TriggerInfoList triggerInfoList;
-	EventInfoList eventInfoList;
-	TriggerStatusType istatusType;
-
+	TriggerStatusType istatus;
```
ashie: I think the prefix "i" isn't needed.

kz0817: I fixed it. Please resume reviewing.

server/src/ArmUtils.cc
```diff
@@ -0,0 +1,250 @@
...
+		if (armTrigger.status != TRIGGER_STATUS_UNKNOWN)
+			obj->createEvent(armTrigger, arg.eventInfoList);
+
+		updateLowerTriggers(arg);
+		updateUpperTriggers(arg);
+	}
+
+	void updateLowerTriggers(UpdateTriggerArg &arg)
+	{
+		for (size_t i = arg.triggerIdx + 1; i < numArmTriggers; i++) {
```
ashie: The range of the "i" seems same with updateUpperTriggers().
Is it correct?

I think the correct code for it is:
```c++
for (size_t i = 0; i < arg.triggerIdx; i++) {
```

kz0817: Thank you for the comment. I fixed it and added a unit test.

ashie:
Confirmed:

* I got a same result with the previous implementation on stalling a Zabbix server (both arm & hap).
* Also I got a same result on reviving them.

ashie:
BTW I found a strange behaviour while testing this modification.
When I filter triggers by the hostname of the self-monitiring, 2 set of self-monitoring triggers are shown.

It's not a bug in this modification.
I'll report it as another issue.

ashie:
> I'll report it as another issue.

#1046

PR #1049, opened 2015-03-03
Save host IDs as a string.

kz0817:
These patches changes the type of the host ID in the DB.
This is one of the pull request in order to save UUID
OpenStack uses for various resources.


kz0817:
Mmmm. A test fails. I'll work on the investigation of it.

> Failure: testSmartBuffer::test_addTooLongString (Over)

kz0817:
The failure reason is that wrong commit. 867f has been contained.
I did rebased for this branch to remove the commit.  And I sent the new pull request (#1050).

PR #1050, opened 2015-03-03
Save host IDs as a string (2nd change)

kz0817:
These patches changes the type of the host ID in the DB.
This is one of the pull request in order to save UUID
OpenStack uses for various resources.


kz0817:
This is a fixed version of #1049.

server/src/DBTablesAction.cc
```diff
@@ -1114,7 +1115,7 @@ string ActionsQueryOption::Impl::makeConditionTemplate(void)
 	// host_id;
 	const ColumnDef &colDefHostId = COLUMN_DEF_ACTIONS[IDX_ACTIONS_HOST_ID];
 	cond += StringUtils::sprintf(
-	  "((%s IS NULL) OR (%s=%%" PRIu64 "))",
+	  "((%s IS NULL) OR (%s=%%s))",
```
ashie: In this function primitive format specifiers are used instead of FMT_* macros.
I think we should use FMT_* macros.

kz0817: No, here '%s' should be used because, because 
corresponding argument at ActionsQueryOption::getCondition()
is rhs(triggerInfo.hostIdInServer) whose return type is const char *.

ashie: I mean we should use `"%" FMT_LOCAL_HOST_ID` instead of `"%s"`.
(`"%" FMT_LOCAL_HOST_ID` is equivalent with `"%s"` but more meaningful.)
And the around code should use FMT_* macros too.

ashie: > corresponding argument at ActionsQueryOption::getCondition()
> is rhs(triggerInfo.hostIdInServer) whose return type is const char *.

`FMT_LOCAL_HOST_ID` shows the corresponding value so that we don't need find it like that.

kz0817: I think we should use the next convention.

When using LocalHostId object: localhostId.
printf("%" FMT_LOCLAL_HOST_ID, localhostId.c_str());

When wrapping DBTermCStringProvider: rhs
printf("%s", rhs(localhostId));

The latter way has the advantages that we can omit c_str()
and replace the type of 'hocalhostId' w/o other changes
because rhs() has some overload versions.
This is one of the reasons why DBTermCStringProvider has been introduced.

Now, the line you pointed out is the latter case. So I use '%s' not  FMT_LOCAL_HOST_ID.

server/common/ZabbixAPI.cc
```diff
@@ -1355,7 +1355,7 @@ void ZabbixAPI::pushTriggersHostid(JSONParser &parser,
 	} else  {
 		for (int i = 0; i < numElem; i++) {
 			startElement(parser, i);
-			pushUint64(parser, itemGroup, "hostid", itemId);
+			pushString(parser, itemGroup, "hostid", itemId);
```
ashie: The dummy data in another side of this block should also be replaced with the string type.

```diff
diff --git a/server/common/ZabbixAPI.cc b/server/common/ZabbixAPI.cc
index 221e384..152418b 100644
--- a/server/common/ZabbixAPI.cc
+++ b/server/common/ZabbixAPI.cc
@@ -1350,7 +1350,7 @@ void ZabbixAPI::pushTriggersHostid(JSONParser &parser,
        startObject(parser, "hosts");
        int numElem = parser.countElements();
        if (numElem == 0) {
-               const uint64_t dummyData = 0;
+               string dummyData;
                itemGroup->addNewItem(itemId, dummyData, ITEM_DATA_NULL);
        } else  {
                for (int i = 0; i < numElem; i++) {
```

Otherwise updating triggers will fail:

```
[DBG] <HatoholException.cc:49> HatoholException: <ItemTable.cc:72> ASSERTION failed: [*groupType0 == *groupType1] : ItemGroupTypes unmatched
[ERR] <ArmZabbixAPI.cc:262> Error on update: <ItemTable.cc:72> ASSERTION failed: [*groupType0 == *groupType1] : ItemGroupTypes unmatched
 2
```

server/src/DBTablesAction.cc
ashie: Ah, I misuderstood it.
rhs(value) always returns `const char*` so the receiver should use '%s".

ashie:
Issue #1046 seems resolved when I use this branch.

ashie:
My review has been done.
It's almost O.K. except following 2 things:

* It fails to get triggers without above patch on my environment.
* Sorry, it conflicts with the master. Please rebase it.

kz0817:
Thanks @ashie san. I'll fix them.

ashie:
#1054

PR #1054, opened 2015-03-04
Save host IDs as a string (3rd challenge)

kz0817:
I fixed the issues commented out in #1050.


ashie:
Confirmed:

* Updating DB fails with the following message (expected behaviour):
```
[CRIT] <main.cc:175> Failed to create a database object. This program is aborted. Reason: <DBTables.cc:293> Invalid Major version: DB/Prog: 00.10 [00]/01.00 [00] (tablesId: 32)
```
* Can build on Ubuntu 14.04 & Scientific Linux 6.3
* It seems fetch all monitoring resources correctly.
* Host filters on all pages seem work correctly.

kz0817:
:blush: Thank you.

PR #1064, opened 2015-03-06
[server] Change the type of Trigger ID to a string.

kz0817:
This change is also for supporting a variety of monitoring systems including OpenStack's ceilometer.

PR #1072, opened 2015-03-09
[server] Change the type of Trigger ID to a string.

kz0817:
This is a rebased version of the  pull request #1064.
The stalls in #1064 probably have been caused by the some invalid memory accesses
that had fixed and merged into master.

### DIFF
--- a/server/common/HatoholArmPluginInterface.h
+++ b/server/common/HatoholArmPluginInterface.h
@@ -208,7 +208,8 @@ struct HapiArmInfo {
 } __attribute__((__packed__));
 
 struct HapiParamTimeOfLastEvent {
-	uint64_t triggerId;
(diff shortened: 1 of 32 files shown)


server/common/HatoholArmPluginInterface.h
```diff
@@ -208,7 +208,8 @@ struct HapiArmInfo {
 } __attribute__((__packed__));
 
 struct HapiParamTimeOfLastEvent {
-	uint64_t triggerId;
+	uint16_t triggerIdLength; // Not include the NULL terminator
+	uint16_t triggerIdOffset; // from the top of this structure
 } __attribute__((__packed__));
```
ashie: Memo: 15.03 breaks not only DB's compatibility but also plugin's compatibility.

ashie:
It seems work fine.

PR #1074, opened 2015-03-10
[HatoholException] Add a feature to call abort() immediately on an exception.

kz0817:
Sometime developers want the detailed stack trace when an exception is
thrown. If the following environment variable is set, hatohol
calls abort() soon after the HatoholException object is created.

  HATOHOL_EXCEPTION_ABORT=1

### DIFF
--- a/server/common/HatoholException.cc
+++ b/server/common/HatoholException.cc
@@ -24,6 +24,7 @@ using namespace std;
 using namespace mlpl;
 
 bool HatoholException::m_saveStackTrace = false;
+bool HatoholException::m_abort = false;
 const int HatoholException::UNKNOWN_LINE_NUMBER = -1;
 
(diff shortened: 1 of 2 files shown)


server/common/HatoholException.cc
```diff
@@ -31,11 +32,20 @@ const int HatoholException::UNKNOWN_LINE_NUMBER = -1;
...
-	if (string("1") == env)
-		m_saveStackTrace = true;
+	struct {
+		void operator ()(const char *envStr, bool &var)
+		{
+			var = false;
+			char *env = getenv(envStr);
+			if (!env)
+				return;
+			if (string(env) =="1")
```
kz0817: I found the careless miss.

kz0817:
I close this PR once.

PR #1086, opened 2015-03-11
Event id to string 2

kz0817:
Some monitoring systems such as OpenStack's ceilometer
handle events as UUID. In order to save that kind of
events, hatohol should also handle them as a string.

### DIFF
--- a/server/common/Utils.h
+++ b/server/common/Utils.h
@@ -259,14 +259,14 @@ public:
 	static void conv(uint64_t &dest, const std::string &src);
 
 	/**
-	 * Calculate addition with the number as a string.
+	 * Sum two variables of which one is a string.
 	 *
 	 * @param num0 A number represented by a string.
 	 * @param num1 An another number.
 	 *
 	 * @return A calculated result.
 	 */
-	static uint64_t add(const std::string &num0, const uint64_t num1);
+	static uint64_t sum(const std::string &num0, const uint64_t num1);
 
 protected:
 	static std::string makeDemangledStackTraceString(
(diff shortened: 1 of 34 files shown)


server/common/Utils.h
```diff
@@ -258,6 +258,16 @@ class Utils {
...
 
+	/**
+	 * Calculate addition with the number as a string.
+	 *
+	 * @param num0 A number represented by a string.
+	 * @param num1 An another number.
+	 *
+	 * @return A calculated result.
+	 */
+	static uint64_t add(const std::string &num0, const uint64_t num1);
```
ashie: I prefer "sum()" instead of "add()" for it.
The former one includes a meaning "calculation".

cosmo0920: Really?
I think that  `sum()` implies a calculation for same type values and `add()` implies a add calculation for arbitrary type values.

ashie: I see, let's go without the change.

kz0817: I take the ashie-san's suggestion.
In the context this method is used, sum() seems to be natural.
I think 'add()' is typically used as a non-static method such as
<pre>
X obj = x;
obj.add(5);
</pre>

cosmo0920: Hmm, I see.


> I think 'add()' is typically used as a non-static method

Ah, that's right.
If so, I have no objection.

ashie:
Will you correct HapProcessCeilometer later (in another pull request)?

kz0817:
| Will you correct HapProcessCeilometer later (in another pull request)?

Yes, I will.

kz0817:
>> Will you correct HapProcessCeilometer later (in another pull request)?
> Yes, I will.

Wakabayashi-san is now creating a patch for the problem.
He will send a PR soon after this PR is merged.

ashie:
It works fine on my environment.

kz0817:
Thanks
:relaxed:

PR #1087, opened 2015-03-11
[HatoholArmPluginGate] Save MonitoringServerStatus on getting items

ashie:
In addition I changed the first argument of
DBTablesMonitoring::addMonitoringServerStatus() from pointer to
reference because there is no possibility to pass a NULL pointer.


ashie:
Thanks!

PR #1089, opened 2015-03-11
[client] Add missing trailing semicolons

cosmo0920:
Missing trailing semicolons causes curious problem when JavaScript minifying.

This curious problems only occurs after minifying JavaScript.


cosmo0920:
Thanks!

PR #1092, opened 2015-03-11
Host group ID in Action condition => String

kz0817:
[server] Change the type of the Hostgroup ID in Action conditionItem ID to a string.

Some monitoring systems such as OpenStack's ceilometer
use UUIDs to identify some resources. In order to save that
kind of data, Hatohol should also handle them as a string.

### DIFF
--- a/server/common/ZabbixAPI.cc
+++ b/server/common/ZabbixAPI.cc
@@ -1364,24 +1364,12 @@ void ZabbixAPI::pushSomethingId(
 	parser.endObject();
 }
 
-void ZabbixAPI::pushTriggersHostid(JSONParser &parser,
-                                   ItemGroup *itemGroup)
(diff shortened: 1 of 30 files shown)


server/common/ZabbixAPI.cc
```diff
@@ -1363,23 +1364,19 @@ void ZabbixAPI::pushTriggersHostid(JSONParser &parser,
 	parser.endObject();
 }
 
+void ZabbixAPI::pushTriggersHostid(JSONParser &parser, ItemGroup *itemGroup)
```
ashie: Using lower case for "id" is mistake?

kz0817: > Using lower case for "id" is mistake?

Yes, It's my mistake. I'll fix it soon.

kz0817: OK. I fixed it. Please continue to review it.

ashie:
It works fine (maybe).

PR #1106, opened 2015-03-13
[DBTablesHost] Remove an unsed struct "Host"

ashie:


ashie:
BTW what does mean "ServerHostDef"?
I can't understand difference between "Host" and "ServerHostDef".
If "Host" isn't really necessary, I want to rename "ServerHostDef" to "Host".

kz0817:
> BTW what does mean "ServerHostDef"?
> I can't understand difference between "Host" and "ServerHostDef".
> If "Host" isn't really necessary, I want to rename "ServerHostDef" to "Host".

Host, that has deleted by the merge, has originally been added in order to
have the data for the record of host_list.
However, it's not been used. So I agree with your suggestion.

PR #1109, opened 2015-03-13
Take some trigger information in events when the save

kz0817:
Until now, some members in EventInfo such as severity and brief
have been filled using a join operation with the trigger table,
when it's retrieved. This way has the following two demerits.

(1) It's hard to delete records in the trigger table.
(2) The number of the save of an event is only once, however
    it may be read many times with join operations.

So this patch resolves the problems by storing duplicated information
 in the event table.

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013-2014 Project Hatohol
+ * Copyright (C) 2013-2015 Project Hatohol
  *
  * This file is part of Hatohol.
  *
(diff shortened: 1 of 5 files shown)


server/src/DBTablesMonitoring.cc
```diff
@@ -1854,16 +1854,12 @@ HatoholError DBTablesMonitoring::getEventInfoList(
 	builder.add(IDX_EVENTS_HOST_NAME);
 	builder.add(IDX_EVENTS_BRIEF);
 
+	// TODO: CONSIDER:  Should we also have extended_info in the event
+	// table ? Then we can delte the following complicated join.
```
ashie: I think so.

kz0817:
> I think so.

@cosmo0920 さんあたりに、さくっと（半日)ぐらいでやってもらうことできますか？

cosmo0920:
はっ、呼ばれてしまいました。

やってみますー。

PR #1115, opened 2015-03-16
[Wercker] building rpm on CentOS 6.x series

cosmo0920:
It is reasonable to confirm building rpm on Wercker.


cosmo0920:
Спасибо!(すぱしーば)

PR #1118, opened 2015-03-16
[server][DBTablesMonitoring] Use events.extended_info instead of triggers' one

cosmo0920:
This work is for https://github.com/project-hatohol/hatohol/pull/1109#issuecomment-81364017.

* These patches keep migration code as possible. 
* Stop to use joining triggers and events tables.


cosmo0920:
Merci beaucoup!

PR #1131, opened 2015-03-18
[RelaseNote] Add items involved with yamato's work.

kz0817:


kz0817:
Oh typooo !
management -> mamage.
I'll send it again after the fix.

PR #1139, opened 2015-03-18
[server][DBTablesMonitoring] Fix exception when filtering with hostname in event page (for releng_15.03)

cosmo0920:
* fix a bug that exception is raised when filtering with hostname in event page
* Add a reproduce test case for this bug


cosmo0920:
Thanks!

PR #1156, opened 2015-03-24
15.03 releng/releasenote ashie 2

ashie:
* Add an item for issue #1151 
* Fold long lines


Mnakagawa:
> - Fold long lines

Thank you for your fix.

PR #1182, opened 2015-04-01
Change from 15.03

Mnakagawa:
We release Hatohol-15.03 yesterday.
This pull request is change of only 15.03 releng branch.


cosmo0920:
:+1:

PR #1185, opened 2015-04-03
Fix version to 15.06 development

Mnakagawa:


kz0817:
Is the failure on Wercker due to either this fix or a Wercker's problem ?
Do anyone know that ?

cosmo0920:
It is Wercker's problem.
This is caused by pulling Docker image phase on Wercker.

kz0817:
Thank you for the reply. I'll merge this PR.

PR #1186, opened 2015-04-04
[wercker][expermental] Use DevToolSet-2 on CentOS 6.x series

cosmo0920:
CentOS 6.x can use DevToolSet version 2.
Thus, gcc 4.8.x series work fine in CentOS 6.x series.

This pull request intends to prepare to use C++11 and partially C++14 feature in Wercker.

=>C++11: [N3337](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf)
=>C++14: [N3797](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf)

* gcc 4.8 seems to support almost all C++11 feature. (see ref: https://gcc.gnu.org/gcc-4.8/cxx0x_status.html)
* gcc 4.8 seems to support C++14 feature partially. (see ref: https://gcc.gnu.org/projects/cxx1y.html)

---
NOTE:
If we want to use almost all C++14 feature in Wercker, we have to build a Docker image which contains gcc 5.x series build toolchain and its runtime.
Or, we have to build a Docker image which contains clang and llvm 3.4 or higher. ref: http://clang.llvm.org/cxx_status.html

### DIFF
--- a/wercker/Dockerfile
+++ b/wercker/Dockerfile
@@ -17,7 +17,6 @@ RUN yum groupinstall -y 'Development tools'
 # setup newer gcc toolchain
 RUN cd /etc/yum.repos.d && wget http://people.centos.org/tru/devtools-2/devtools-2.repo
 RUN yum install -y devtoolset-2-gcc devtoolset-2-binutils devtoolset-2-gcc-c++
-RUN scl enable devtoolset-2 bash
 RUN yum install -y Django
 RUN rpm -Uvh http://sourceforge.net/projects/cutter/files/centos/cutter-release-1.1.0-0.noarch.rpm
 RUN yum install -y cutter tar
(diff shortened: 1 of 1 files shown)


kz0817:
Thank you for the quick action for using C++11/14. :yum:

I have some questions. (Actually I want to check them by myself, but now I have no enough time)
Do you know if there is the same or corresponding package to devtoolset-2 in RHEL?
And the package replaces gcc/binutils, doesn't it ?
Are there required packages when the compiled binaries are running on other than the build machine ?

cosmo0920:
> Do you know if there is the same or corresponding package to devtoolset-2 in RHEL?

I think that RHEL has devtoolset-2 package (same name for CentOS!) which called in `Red Hat Developer Toolset`. ref: https://access.redhat.com/documentation/en-US/Red_Hat_Developer_Toolset/2/html/User_Guide/sect-Red_Hat_Developer_Toolset-Install.html

> And the package replaces gcc/binutils, doesn't it ?

No, it doesn't replace system's gcc/binutils.
They are installed in different location such as `/opt/rh/devtoolset-2/root/`.

cosmo0920:
Thanks!

PR #1187, opened 2015-04-04
[Travis] Switch to use gcc-4.8.x series in Travis

cosmo0920:
Travis CI cpp instance's default gcc is 4.6.
This pull request intents to use gcc-4.8.x series in Travis CI. 

Related to #1186.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,12 +34,16 @@ matrix:
     - compiler: gcc
       env: RUN_TEST="client/test/python/run-test.sh"
 
+before_install:
+  - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+  - sudo apt-get update -qq
+  - if [ "$CXX" = "clang++" ]; then sudo apt-get install -qq libstdc++-4.8-dev; fi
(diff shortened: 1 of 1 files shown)


kz0817:
Before merging this patch, we want to summarize the impact of using new GCC with C++11 features as a default especially on CentOS6/7 at runtime.

Do we need to install additional libraries ?

cosmo0920:
> Do we need to install additional libraries ?

Yes, we do. But we need to do for a few old environment.

I think that Ubuntu 12.04 LTS (used by Travis CI) and CentOS 6.x are needed additional runtime library for gcc-4.8.
We need to install additional libraries which are `libstdc++4.8-dev` in Ubuntu Precise (12.04 LTS).
We need to install additional libraries which are `devtoolset-2-runtime` in CentOS 6.x.

CentOS 7.x's and Ubuntu Trusty(14.04 LTS)'s default is gcc 4.8. 
So, we needn't to do anything, though.

kz0817:
Thank you for the reply.
I'll merge the Pull Request.

cosmo0920:
Thanks!

PR #1192, opened 2015-04-14
Update CodingConvension.md

cosmo0920:
I've forgot to update this file.... :confounded: 
We start to use C++11/C++14 (only supported by g++-4.8 series), doesn't we?

NOTE:
g++-4.8 supports only C++14 features which is `Return type deduction for normal functions` ([N3638](https://isocpp.org/files/papers/N3638.html)).
Otherwise, C++14 feature support in g++-4.8 is partially or nothing.

### DIFF
--- a/CodingConvention.md
+++ b/CodingConvention.md
@@ -27,10 +27,8 @@ An input parameter shall be 'const TYPE &'.
 
 #### Not using
 - Boost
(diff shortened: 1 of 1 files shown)


cosmo0920:
Thanks!

Issue #1144, opened 2015-03-19
[Hatohol] Shouldn't return the HTTP status "401"


ashie:
Because it's for HTTP authentication, the response must include a WWW-Authenticate header field. 

http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2

But Hatohol server doesn't use HTTP authentication.
It was itroduced at #1030.

ashie:
Fixed in #1194.

PR #1194, opened 2015-04-22
[server][FaceRest] Use 400 status when authentcation failed

cosmo0920:
This work is related to #1144.
Hatohol Server does not use HTTP authentication, so it should be used Bad Request (400) status code instead of Unauthorized(401).


cosmo0920:
Thanks!

PR #1198, opened 2015-04-28
Use extendedInfo in TriggerInfo & EventInfo directly

ashie:
Because we decided to accept any string in it:

https://github.com/project-hatohol/HAPI-2.0-Specification


cosmo0920:
:+1:

ashie:
Thanks!

PR #1200, opened 2015-05-08
[configure][server] Enable C++11 by default

cosmo0920:
This pull request permits to use C++11 by default.

Related to #1186.


cosmo0920:
Thanks!

PR #1196, opened 2015-04-28
Add extended info for monitoring server

ashie:
Because it's required in HAPI 2.0:

https://github.com/project-hatohol/HAPI-2.0-Specification

PR #1201, opened 2015-05-11
Add extended info for monitoring server

ashie:
Because it's required in HAPI 2.0:

https://github.com/project-hatohol/HAPI-2.0-Specification

In addition change the length of servers.base_url to 2047 according to https://github.com/project-hatohol/HAPI-2.0-Specification/issues/5

Revise #1196.


ashie:
Thanks!

PR #1202, opened 2015-05-13
[server][DBTables][LastInfo] Implement LastInfo

cosmo0920:
Because it's required in HAPI 2.0:
https://github.com/project-hatohol/HAPI-2.0-Specification#user-content-getlastinfomethod.


server/common/Params.h
```diff
@@ -89,6 +90,12 @@ typedef int IncidentTrackerIdType;
 typedef uint64_t ActionLogIdType;
 #define FMT_ACTION_LOG_ID PRIu64
 
+typedef uint64_t LastInfoIdType;
```
kz0817: Why is this type is defined as 'uint64_t', although the corresponding column type in COLUMN_DEF_LAST_INFO is INT ?

server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,483 @@
...
+#include "DBTablesMonitoring.h"
+#include "DBTablesLastInfo.h"
+#include "Mutex.h"
+#include "ItemGroupStream.h"
+#include "UnifiedDataStore.h"
+#include "DBTermCStringProvider.h"
+using namespace std;
+using namespace mlpl;
+
+const char *TABLE_NAME_LAST_INFO     = "last_info";
```
kz0817: Meaningless spaces before '='.

server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,483 @@
...
+		updateDB,
+	};
+	return setupInfo;
+};
+
+// ---------------------------------------------------------------------------
+HatoholError DBTablesLastInfo::checkPrivilegeForAdd(
+  const OperationPrivilege &privilege, const LastInfoDef &actionDef)
+{
+	UserIdType userId = privilege.getUserId();
```
kz0817: I like to use 'const' for the read only variable.
Of course, there is no big problem if it doesn't

server/common/Params.h
```diff
@@ -89,6 +90,12 @@ typedef int IncidentTrackerIdType;
...
+#define FMT_LAST_INFO_ID PRIu64
+
+typedef std::string LastInfoServerIdType;
```
kz0817: Why the type is string, not ServerIdType (int) ?

server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,483 @@
...
+	return HTERR_OK;
+}
+
+static string makeIdListCondition(const LastInfoIdList &idList)
+{
+	string condition;
+	const ColumnDef &colId = COLUMN_DEF_LAST_INFO[IDX_LAST_INFO_ID];
+	condition = StringUtils::sprintf("%s in (", colId.columnName);
+	LastInfoIdListConstIterator it = idList.begin();
+	while (true) {
```
kz0817: For this situation, maple::SeparatorInjector is prepared.

The below is from src/DBAgentMySQL.cc
<pre>
        SeparatorInjector commaInjector(",");
        string query = StringUtils::sprintf("INSERT INTO %s (",
                                            insertArg.tableProfile.name);
        for (size_t i = 0; i < numColumns; i++) {
                const ColumnDef &columnDef = insertArg.tableProfile.columnDefs[i];
                commaInjector(query);
                query += columnDef.columnName;
        }
</pre>

server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,483 @@
...
+	arg.add(IDX_LAST_INFO_DATA_TYPE);
+	arg.add(IDX_LAST_INFO_VALUE);
+	arg.add(IDX_LAST_INFO_SERVER_ID);
+
+	// condition
+	arg.condition = option.getCondition();
+
+	getDBAgent().runTransaction(arg);
+
+	const ItemGroupList &grpList = arg.dataTable->getItemGroupList();
```
kz0817: This comment is just an impression.
Since we accept C++11, I'd like to rewrite this pattern used in Hatohol many times with foreach-like API of itemGrpItr and a lambda equation.

server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,483 @@
...
+		{
+			dbAgent.deleteRows(arg);
+			numAffectedRows = dbAgent.getNumberOfAffectedRows();
+		}
+	} trx;
+	trx.arg.condition = makeConditionForDelete(idList, privilege);
+	getDBAgent().runTransaction(trx);
+
+	// Check the result
+	if (trx.numAffectedRows != idList.size()) {
```
kz0817: Is this check really needed ?
How about just showing a warning ?
In the current architecture, the number of the affected should be the same as the number of requested.
However, we could add the mechanism that deletes records concurrently.

kz0817:
I'm finished. Please check the comments.

server/common/Params.h
cosmo0920: I've fixed the corresponding column (last_info_id) type and length: https://github.com/project-hatohol/hatohol/commit/06101c718e2fad7a66c235f5e631916f2511d9ca.

server/common/Params.h
cosmo0920: OK, I'll use int type for LastInfoServerIdType.

cosmo0920:
I fixed some reviewed problems.
But a few thing is still left.
I'll try to resolve them later.

server/src/DBTablesLastInfo.cc
cosmo0920: Although, I think that I want to check result here.

In DBTablesAction, similar check exists.
In DBTablesConfig, deleting result check does not exist.
Should we always check deleting result, or omit this kind of checking?

cosmo0920:
I've fixed a few pointed out things and add comment.

cosmo0920:
Following patch is just an impression.  It should not be take care.

In C++11, added `constexpr` keywords.
Following use case seems to be useful?

```diff
diff --git a/server/src/DBTablesLastInfo.cc b/server/src/DBTablesLastInfo.cc
index 5d0b54e..7b2f0cc 100644
--- a/server/src/DBTablesLastInfo.cc
+++ b/server/src/DBTablesLastInfo.cc
@@ -32,7 +32,7 @@
 using namespace std;
 using namespace mlpl;
 
-const char *TABLE_NAME_LAST_INFO = "last_info";
+constexpr const char *TABLE_NAME_LAST_INFO = "last_info";
 
 const int DBTablesLastInfo::LAST_INFO_DB_VERSION =
   DBTables::Version::getPackedVer(0, 1, 0);
@@ -43,7 +43,7 @@ static void operator>>(
        lastInfoType = itemGroupStream.read<int, LastInfoType>();
 }
 
-static const ColumnDef COLUMN_DEF_LAST_INFO[] = {
+constexpr static const ColumnDef COLUMN_DEF_LAST_INFO[] = {
 {
        "last_info_id",                    // columnName
        SQL_COLUMN_TYPE_BIGUINT,           // type
diff --git a/server/src/DBTablesLastInfo.h b/server/src/DBTablesLastInfo.h
+++ b/server/src/DBTablesLastInfo.h
@@ -25,8 +25,8 @@
 #include "DBTables.h"
 #include "Params.h"
 
-const static LastInfoIdType INVALID_LAST_INFO_ID = -1;
-const static LastInfoServerIdType INVALID_LAST_INFO_SERVER_ID = -1;
+constexpr const static LastInfoIdType INVALID_LAST_INFO_ID = -1;
+constexpr const static LastInfoServerIdType INVALID_LAST_INFO_SERVER_ID = -1;
 
 enum LastInfoType {
        LAST_INFO_ALL = -1,
```

cosmo0920:
Oh, I've forgot to rebase with #1200.
So, I close this pull request.

server/src/DBTablesLastInfo.cc
cosmo0920: I've found a use case `range-based for` for this pattern like this:

```cpp
	const ItemGroupList &grpList = arg.dataTable->getItemGroupList();
	for (auto item : grpList) {
		ItemGroupStream itemGroupStream(item);
		LastInfoDef lastInfoDef;

		itemGroupStream >> lastInfoDef.id;
		itemGroupStream >> lastInfoDef.dataType;
		itemGroupStream >> lastInfoDef.value;
		itemGroupStream >> lastInfoDef.serverId;

		lastInfoDefList.push_back(lastInfoDef);
	}
```

We should rewrite this pattern by more sophisticated way near future?

PR #1205, opened 2015-05-14
[travis] Try to avoid unexpected failure

ashie:
* Avoid using parallel build to reduce load
* Join some tests to reduce number of builds

https://travis-ci.org/project-hatohol/hatohol/builds/62514289

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ compiler:
 env:
   global:
     - MLPL_LOGGER_LEVEL=WARN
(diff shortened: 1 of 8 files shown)


ashie:
Thanks!

kz0817:
Although it is supposed not to related with this patch, although I'm concerned that the error in the Travis: https://travis-ci.org/project-hatohol/hatohol/jobs/62519287. This indicates that there may be a potential problem.

PR #1206, opened 2015-05-15
[server][DBTables][LastInfo] Implement LastInfo2

cosmo0920:
Resend #1202.


server/src/DBTablesLastInfo.cc
```diff
@@ -0,0 +1,484 @@
...
+}
+
+static string makeIdListCondition(const LastInfoIdList &idList)
+{
+	string condition;
+	const ColumnDef &colId = COLUMN_DEF_LAST_INFO[IDX_LAST_INFO_ID];
+	SeparatorInjector commaInjector(",");
+	condition = StringUtils::sprintf("%s in (", colId.columnName);
+	LastInfoIdListConstIterator it = idList.begin();
+	while (true) {
```
kz0817: This block can be written with few lines simply as
<pre>
for (; it != idList.end(); ++it) {
    const int &id = *it;
    commnadInjector(condition);
    condition += StringUtils::sprintf("%d", id); 
}

or

for (; it != idList.end(); ++it) {
    commnadInjector(condition);
    condition += StringUtils::sprintf("%d", *it);
}

</pre>

kz0817:
Review had been done. I wrote one comment.

server/src/DBTablesLastInfo.cc
cosmo0920: I like later one style.

But, we can use more sophisticated range-based for style:

```cpp
for (auto id : idList) {
	commaInjector(condition);
	condition += StringUtils::sprintf("%ld", id);
}
```

How about this one?

cosmo0920:
I've fixed a reviewed issue.

server/src/DBTablesLastInfo.cc
kz0817: Yes, I want to use *type* inference feature.

2015-05-19 11:42 GMT+09:00 Hiroshi Hatake <notifications@github.com>:

kz0817:
I merge this PR.

BTW, recently the build on TravisCI often failed with internal error.
If the reason is the short of a memory, it is worth reducing the number of parallel build.

cosmo0920:
Thanks!

PR #1207, opened 2015-05-15
[AMQPConnection] Add an ability to publish a message

ashie:
AMQPConnection (which is a wrapper class for rabbitmq-c) doesn't have an ability to send a message.
@cosmo0920 and I added an ad-hoc implementation of it.
Although I'll heavily modify it later to introduce an abstraction layer of the communication,
I want to merge the current code to implement the skeleton of the server side of HAPI-2.0 ASAP.

### DIFF
--- a/server/src/AMQPConnectionInfo.cc
+++ b/server/src/AMQPConnectionInfo.cc
@@ -17,6 +17,8 @@
  * <http://www.gnu.org/licenses/>.
  */
 
+#include <string.h>
+#include <stdlib.h>
 #include <string>
 #include <amqp.h>
 #include <Logger.h>
@@ -32,6 +34,7 @@ using namespace mlpl;
 struct AMQPConnectionInfo::Impl {
 	Impl()
 	: m_URL(),
+	  m_URLBuf(NULL),
 	  m_parsedURL(),
 	  m_queueName(),
 	  m_timeout(DEFAULT_TIMEOUT)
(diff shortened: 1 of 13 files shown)


cosmo0920:
:green_heart:

server/src/AMQPConnectionInfo.cc
```diff
@@ -32,6 +34,7 @@ using namespace mlpl;
 struct AMQPConnectionInfo::Impl {
 	Impl()
 	: m_URL(),
+	  m_URLBuf(NULL),
```
kz0817: m_URLBuf is only used in setURL(), isn't it.
If so, it's unnecessary to be an instance variable.
Or will you use it in the methods you are going to make later.

ashie: > m_URLBuf is only used in setURL(), isn't it.
> If so, it's unnecessary to be an instance variable.

No, it's used by m_parsedURL.

It's a labyrinth of rabbitmq-c.
This buffer is referred from m_parsedURL (amqp_connection_info).
So we need to keep it until m_parsedURL is destroyed.

ashie: amqp_parse_url(url, connectionInfo) do the following things:

* Replace some characters in url with "\0" to split the url string to url elements
* Store pointer to each url elements into connectionInfo. It doesn't copy the strings, refer them directly.

kz0817: OK. I see.
But we want to recognize the relationship easily.
Anyway,  I'll merge this PR.

ashie: > But we want to recognize the relationship easily.

I agree with you.
I'll add a comment or something in another PR.

PR #1209, opened 2015-05-19
[testAMQPConnection] Fix a bug that a test isn't omitted

ashie:
test_transferMessage should be executed only when TEST_AMQP_URL
is set, but it ignores the env value.


ashie:
I merge this PR by myself because it's an easy fix.

PR #1210, opened 2015-05-21
Add "uuid" column to server_types and arm_plugins

ashie:
Add "uuid" column which is required in HAPI 2.0 to server_types & arm_plugins tables.


ashie:
Now I've noticed a fatal issue.
It can't register two or more plugins because "type" is duplicate.
I'll fix it and send another PR.

PR #1212, opened 2015-05-21
Add "uuid" column to server_types and arm_plugins

ashie:
Add "uuid" column which is required in HAPI 2.0 to server_types & arm_plugins tables.

To distinguish old plug-ins & new plug-ins, I introduced multiple column index to server_types table.


ashie:
Thanks!

PR #1217, opened 2015-05-25
[server] Remove a needless template "Signal2" and related codes

ashie:
Because it's completely same with Signal0 except the name.


ashie:
Thank you!

PR #1220, opened 2015-05-26
[RestResourceServer] Remove needless calllback for fetching triggers

ashie:
Since handlerTriggerUpdateServer() sets HTTP response immediately
in the function, it shouldn't do it again in the callback function.
So the callback function isn't needed anymore because it doesn't
do other things.


ashie:
I'll add issue number to the commit message.

PR #1222, opened 2015-05-27
[server][DBTablesLastInfo] Revise getItem from DB pattern with range-based for

cosmo0920:
I didn't notice this use case in #1202 and #1206.

We should rewrite this pattern by more sophisticated way near future?

### DIFF
--- a/server/src/DBTablesLastInfo.cc
+++ b/server/src/DBTablesLastInfo.cc
@@ -207,9 +207,8 @@ HatoholError DBTablesLastInfo::getLastInfoList(LastInfoDefList &lastInfoDefList,
 	getDBAgent().runTransaction(arg);
 
 	const ItemGroupList &grpList = arg.dataTable->getItemGroupList();
-	ItemGroupListConstIterator itemGrpItr = grpList.begin();
-	for (; itemGrpItr != grpList.end(); ++itemGrpItr) {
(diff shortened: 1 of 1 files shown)


server/src/DBTablesLastInfo.cc
```diff
@@ -207,9 +207,8 @@ HatoholError DBTablesLastInfo::getLastInfoList(LastInfoDefList &lastInfoDefList,
 	getDBAgent().runTransaction(arg);
 
 	const ItemGroupList &grpList = arg.dataTable->getItemGroupList();
-	ItemGroupListConstIterator itemGrpItr = grpList.begin();
-	for (; itemGrpItr != grpList.end(); ++itemGrpItr) {
-		ItemGroupStream itemGroupStream(*itemGrpItr);
+	for (auto item : grpList) {
```
kz0817: I propose 'itemGrp' as a variable name, because it is not a single item but a set of items.

kz0817: Sorry. I was confused. The previous comment is probably wrong. Don't mind.

cosmo0920: Oh, I see. Yes, `range_declalation`is not a single item. They contain multiple values.

cosmo0920:
Thanks!

PR #1232, opened 2015-06-10
Initial implementation of HatoholArmPluginGateHAPI2

ashie:
Although it's not completed yet, I want to merge it once because it's already too big...

Known issues:

* Error handling is too weak
* "fetchEvent" isn't implemented yet
* "fetchItems" & "fetchTriggers" don't accept "hostIds" parameter
* "mayMore" parameter isn't handled yet
* "updateHostParent" doesn't support updateType=ALL
* And so on...

I believe it doesn't break existing functions.


ashie:
Wow! Thanks!

kz0817:
I'm so tired.
I tried to feel abnormal smell in the suggested code
because it's hard to follow all the methods and variables deeply.

PR #1235, opened 2015-06-11
[HAPI 2.0] Fix some broken validations

ashie:
And add some error messages.

### DIFF
--- a/server/common/HatoholArmPluginInterfaceHAPI2.cc
+++ b/server/common/HatoholArmPluginInterfaceHAPI2.cc
@@ -110,8 +110,8 @@ struct JsonRpcObject {
 		JSONParser::ValueType type = parser.getValueType("method");
 		if (type != JSONParser::VALUE_TYPE_STRING) {
 			m_errorMessage =
-				"Invalid request: "
-				"Invalid type for \"method\"!";
+			  "Invalid request: "
+			  "Invalid type for \"method\"!";
 			return;
 		}
 		parser.read("method", m_methodName);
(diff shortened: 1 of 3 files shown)


server/common/HatoholArmPluginInterfaceHAPI2.cc
```diff
@@ -46,81 +46,39 @@ std::list<HAPI2ProcedureDef> defaultValidProcedureList = {
...
-  : public AMQPMessageHandler
-{
-public:
-	AMQPHAPI2MessageHandler(HatoholArmPluginInterfaceHAPI2 &hapi2)
-	: m_hapi2(hapi2)
-	{
-	}
-
-	enum class JsonRpcObjectType {
+struct JsonRpcObject {
```
kz0817: In other code, we use capital letters for abbreviated words such as JSON.
Here, we should use JSON_RPCObject or JSON_RPC_Object ?
But I feel a little uncomfortable...
JsonRpcObject looks simpler than JSON_RPCObject...

ashie: > In other code, we use capital letters for abbreviated words such as JSON.

Yes, using "JSON" is our policy.
But I held a the same impressions as you:

> Here, we should use JSON_RPCObject or JSON_RPC_Object ?
> But I feel a little uncomfortable...
> JsonRpcObject looks simpler than JSON_RPCObject...

So, I used "JsonRpcObject".

But now I think we shouldn't change our policy.
I'll replace it with JSONRPCObject.

ashie:
> I'll replace it with JSONRPCObject.

I did it.
Please continue your review.

ashie:
Thanks!

PR #1252, opened 2015-06-16
[testHatoholArmPluginGateHAPI2] Fix broken tests in #1250

ashie:
loadTestDBServer() fails on saving HAPI2 server info because
it doesn't pass ArmPluginInfo to DBTablesConfig::addTargetServer().


ashie:
I forgot to remove a needless data.

PR #1263, opened 2015-06-18
Add a basic mechanism for WebUI plugins

ashie:
* Plugin files are loaded from /static/js.plugins
* The file names should be {Plugins's UUID}.js
* A plugin must define a new object hatohol.hap_{UUID}
  * Obsolete types (zabbix, nagios, ...) use "type" value instead of UUID
* The plugin implementation should be added to the object.
  * type: UUID
  * label: string
  * getTopURL()
  * getItemGraphURL()
  * getMapsURL()


ashie:
Thanks!

PR #1264, opened 2015-06-19
[client] Avoid using "valueType" in items

ashie:
Because we are going to remove it in HAPI 2.0.
See also https://github.com/project-hatohol/HAPI-2.0-Specification/issues/43

It also fix a bug that an incorrect "prevValue" is shown for a string item.


ashie:
Thanks!

PR #1275, opened 2015-06-22
Implement a RabbitMQHapiConnector.

Mnakagawa:


server/hap2/test/TestHaplib.py
```diff
@@ -156,3 +159,18 @@ def test_create(self):
...
+    def test_setup(self):
+        port = os.getenv("RABBITMQ_NODE_PORT")
+        amqp_address = os.getenv("RABBITMQ_NODE_ADDRESS")
+        transporter_args = {"direction": transporter.DIR_SEND,
+                            "amqp_broker": amqp_address,
+                            "amqp_port": port,
+                            "amqp_vhost": "test",
+                            "amqp_queue": "test_queue",
+                            "amqp_user": "test_user",
+                            "amqp_password": "test_password"}
```
ashie: Now I have a question about treating "password".
Do these plugins receives passwords via only command line arguments?
If it's true, it's dangerous.
Any user can see it by `ps ax` command.

ashie:
This PR itself is OK, so I merge it.
On the other hand I have a question as I mentioned before.
If it's true, please resolve it in other PR.

server/hap2/test/TestHaplib.py
Mnakagawa: > Do these plugins receives passwords via only command line arguments?

That's right.
Anyway, I will send PR of other developmented.
I'll fix this proiblem after merge all PR.

PR #1277, opened 2015-06-22
Implement a Sender.

Mnakagawa:


Mnakagawa:
I found mistake.
I resend PR after fix this mistake.

PR #1280, opened 2015-06-23
 [client] Fix a bug that server type labels for plugins become "Invalid"

ashie:
Because a UUID isn't passed to makeMonitoringSystemTypeLabel().


ashie:
Thanks!

PR #1284, opened 2015-06-24
[HatoholArmPluginGateHAPI2] Fill some uninitialized members of received resources

ashie:


ashie:
Sorry, I've found one more bug to fix in this PR.

PR #1285, opened 2015-06-24
[HatoholArmPluginGateHAPI2] Fill unintialized members

ashie:
* Fill some uninitialized members
* Fix a bug that hostIdInServer aren't stored


ashie:
Thanks!

PR #1292, opened 2015-06-25
[Issue-945] Remove "Previous value" column in "LatestData" page and DBTable.

t-miyazawa:


kz0817:
This PR breaks the DB compatibility.
On 15.06, we won't be planning the major DB change.
 Using 'alter table' is required.

t-miyazawa:
DBを変更しないような修正方法に変更するため取り下げます。

PR #1294, opened 2015-06-25
Fix a problem that raises an exception in TravisCI test.

kz0817:
For example, https://travis-ci.org/project-hatohol/hatohol/builds/68278198
failed with the following exception.

  File "/usr/lib/python2.7/dist-packages/pika/connection.py", line 56, in __init__

    raise TypeError("Port must be an int")

Since the test of RabbitMQConnector is done in TestRabbitMQConnector,
the test causing the problem is redundant and unnecessary.


ashie:
Thanks!

PR #1301, opened 2015-06-25
Add a utility for quick testing.

kz0817:
This utility sends some RPC request such as fetchTriggers and
fetchItems to look at the behaivor of the Hatohol Plugins quickly.


kz0817:
Thanks.

PR #1306, opened 2015-06-26
[HatoholArmPluginGateHAPI2] Rename "histories" to "samples"

ashie:
Follow the change of the specification:

https://github.com/project-hatohol/HAPI-2.0-Specification/issues/47

I'll also modify REST API in another pull request.


ashie:
Thanks!

PR #1313, opened 2015-06-29
[server][test][HAPI2] Use std::atomic

cosmo0920:
C++11 also provides `std::atomic`.

This class provides 

* `operator=` (same as `AtomicValue<T>.operator=` functionality)
* `operator()` (same as `AtomicValue<T>.operator()` functionality)
* `load()` (same as `AtomicValue<T>.get()` functionality)
* `store()` (same as `AtomicValue<T>.set()` functionality)

and various convenient member functions and operators since C++11.

ref. [cppreference.com | std::atomic](http://en.cppreference.com/w/cpp/atomic/atomic)

### DIFF
--- a/server/test/testAMQPConnection.cc
+++ b/server/test/testAMQPConnection.cc
@@ -19,9 +19,9 @@
 
 #include <gcutter.h>
 #include <cppcutter.h>
+#include <atomic>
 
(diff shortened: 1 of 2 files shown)


cosmo0920:
Thanks!

PR #1317, opened 2015-06-29
[HatoholServerEditDialogParameterized] Enable to select HAPI2 server types

ashie:
Should use "uuid" instead of "type" to distinguish each HAPI2
server type, because the "type" value of HAPI2 server types always
"7" (hatohol.MONITORING_SYSTEM_HAPI2).


ashie:
Thanks!

PR #1329, opened 2015-06-30
[HAPI2] Initial support of starting & stopping HAPI2 plugins automatically

ashie:
If armPluginInfo.path isn't empty, execute it with an argument "start" or "stop" when a instance of
HatoholArmPluginGateHAPI2 starts or stops.

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -323,14 +323,32 @@ struct HatoholArmPluginGateHAPI2::Impl
 		}
 	};
 
+	bool isPluginControlScriptAvailable(void)
+	{
+		if (m_pluginInfo.path.empty())
+			return false;
+		bool executable = g_file_test(m_pluginInfo.path.c_str(),
+					      G_FILE_TEST_IS_EXECUTABLE);
+		if (!executable)
+			MLPL_DBG("The plugin control script isn't executable:"
+				 " %s\n", m_pluginInfo.path.c_str());
+		return executable;
+	}
+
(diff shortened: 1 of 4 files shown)


ashie:
To tell the truth, it's not enough yet.
It doesn't provide a away to find the process to stop yet.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -321,6 +322,104 @@ struct HatoholArmPluginGateHAPI2::Impl
...
 	};
+
+	bool isPluginControlScriptAvailable(void)
+	{
+		if (m_pluginInfo.path.empty())
+			return false;
+		bool executable = g_file_test(m_pluginInfo.path.c_str(),
+					      G_FILE_TEST_IS_EXECUTABLE);
+		if (!executable)
+			MLPL_DBG("The plugin control script isn't executable:"
```
kz0817: The higher message level would be better. For example INFO or WARN.
And users get a chance to notice the cause of problem.

kz0817: > Or, is a message of failure of the execution shown in other place ?

Yes, the failure message is shown in runPluginControlScript. However, the reason isn't.
So here the message including a cause should be logged with more higher level.

ashie: > So here the message including a cause should be logged with more higher level. 

I think so too. I'll change it.

ashie: I fixed it.

ashie:
Thanks!

PR #1333, opened 2015-07-01
[hap2][TestZabbixAPI] Fix failed tests

ashie:
The parameter "priority" in trigger.get of ZabbixAPI should be
a string, not a number.


kz0817:
I check the all tests passed on my local environment on Ubuntu 14.04.

ashie:
Thanks!

PR #1339, opened 2015-07-02
[hap2][CommandQueue] Fix an inappropriate timeout value for Queue.get()

ashie:
The combination of block=True and timeout=0 causes busy loop.
The "timeout" out should be None or a positive value when blocking
process is needed.

https://docs.python.org/2.7/library/queue.html#Queue.Queue.get


ashie:
Thanks!

PR #1344, opened 2015-07-02
[DBTablesConfig][RestResourceServer] Fix an issue that can't find an appropriate HAPI2 server type

ashie:
"uuid" is required on searching a HAPI2 server type


ashie:
Thanks!

PR #1348, opened 2015-07-02
[test][EventsView] Fix broken tests

ashie:
EventsView calls stupidsort() of stupidtable, and it calls
setTimeout(function(){...}, 10) implicitly. It breaks tests of
EvensView.

This modification avoid it by waiting 20 ms.


ashie:
Trivial fix.

kz0817:
:clap:

PR #1371, opened 2015-07-06
Fix required file names

Mnakagawa:


Mnakagawa:
I confirm to make rpm in my environment.

PR #1378, opened 2015-07-06
[HatoholArmPluginGateHAPI2] Fix broken timestamp parse on CentOS 6.5

ashie:
Length modifier of format specifier is invalid.


ashie:
issue #1376

PR #1396, opened 2015-07-07
Terminate all the processes when the main plugin gets SIGTERM.

kz0817:
This patch makes all the HAP processes terminate when SIGTERM is received.


kz0817:
It might be better that only the main plugin handles the signal. I will consider it again.

kz0817:
I send pull request again without any fix.
I think the way of the patch is less troublesome.

ashie:
@Mnakagawa said we should send PR against 15.06_releng/master (https://github.com/project-hatohol/hatohol/pull/1377#issuecomment-119076319).
What do you think about it?

PR #1403, opened 2015-07-08
Fix 1362

kz0817:
This patches enables to show HAP plugin dependent host name using
JS plugin mechanism.


kz0817:
This is request for master. So I once close it.

PR #1406, opened 2015-07-08
[haplib] Make sure there's 'method' before message is sent to RPC queue.

kz0817:
An unexpected message is sometimes sent to MainPlugin.
This patch drops if's that case.


ashie:
Ack

PR #1407, opened 2015-07-08
[HAPI2] Improve launching plug-in control scripts

ashie:
* Pass more convenient environment variables to the plug-in control script
* Run the plug-in control script synchronously (these script shouldn't block long time)
* Complete relative path of plug-in control script
* Handle "passiveMode" parameter from WebUI correctly

I'm now testing plugins control scripts in the following branch:

https://github.com/project-hatohol/hatohol/tree/15.06_releng/plugin_control_script_2

### DIFF
--- a/server/common/Utils.h
+++ b/server/common/Utils.h
@@ -36,8 +36,12 @@
 #endif
 
 #ifndef SOUP_VERSION_2_32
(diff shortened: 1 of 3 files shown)


ashie:
Sorry, I've found a memory leak.

PR #1409, opened 2015-07-08
Ceilometer hostname fixup

kz0817:
When the host ID of the ceilometer's alarm is "N/A", it's shown as Unknown ID.
These patches fixes it.


ashie:
Ack

PR #1475, opened 2015-07-21
[client][servers_view] Rename some variables to represent their actual roles

ashie:
paramArray ->
paramsArrayMap

paramObj ->
paramsArray


ashie:
Trivial fix.

PR #1476, opened 2015-07-21
[client][plugins] Don't escape URL 

ashie:
They should be escaped at caller side.


ashie:
It's travial fix.
And I confirmed that all tests for utils.js passes on my environment.

PR #1477, opened 2015-07-21
[travis] Fix install error of mocha-phantomjs

ashie:
It requires an old versio of phantomjs (1.9.1 - 1.9.7-15).

The error log:
5495 error peerinvalid The package phantomjs does not satisfy its siblings' peerDependencies requirements!
5495 error peerinvalid Peer mocha-phantomjs@3.6.0 wants phantomjs@1.9.1 - 1.9.7-15

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
   - npm install -g mocha
   - npm install -g expect.js
   - npm install -g sinon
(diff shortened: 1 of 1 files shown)


ashie:
I confirmed that mocha-phantomjs is installed correctly.

cosmo0920:
Good stuff!

PR #1478, opened 2015-07-21
[HAP2][WebUI] Implement getting URLs

ashie:
Implement following plugin functions:

* HAP2 Zabbix
    * getTopURL()
    * getMapsURL()
    * getItemGraphURL()
* HAP2 Nagios
    * getTopURL()

PR #1480, opened 2015-07-22
[HAP2][WebUI] Implement getting URLs

ashie:
Implement following plugin functions:

* HAP2 Zabbix
    * getTopURL()
    * getMapsURL()
    * getItemGraphURL()
* HAP2 Nagios
    * getTopURL()

Revise #1478


cosmo0920:
I confirmed that HAP2 Nagios has a strange behavior.
Otherwise works fine.

cosmo0920:
We should use extendInfo in HAP2 Nagios....

ashie:
> I confirmed that HAP2 Nagios has a strange behavior.

Oops.
Sorry, I'll fix it.

PR #1481, opened 2015-07-22
[hap2][client][zabbix] Implement get{Top,Maps,ItemGraph}URL()

ashie:
And simplify getTopURL() in nagios.js.

(Remove problematic codes in #1480)


ashie:
> We should use extendInfo in HAP2 Nagios....

In the current implementation, users are required to enter raw JSON because there is no way to
render input form for it. I want to add a mechanism to do it.

cosmo0920:
Do improving HAP2 nagios implementation is another pull request?

cosmo0920:
Merged!

ashie:
Thanks!

PR #1485, opened 2015-07-23
[HatoholArmPluginGateHAPI2] Implement a missing error response

ashie:
An error response should be returned when it receives a too large
events array.


ashie:
Trivial fix.

PR #1487, opened 2015-07-23
[HatoholArmPluginGateHAPI2] Fix a bug that the self monitoring feature causes crashes

ashie:
Should use a copy of ServerInfo instead of a reference.


ashie:
Trivial fix.

PR #1488, opened 2015-07-23
[AMQPConnection] Correct error handling of AMQPConnection

ashie:
* Always try reconnecting when an error occured
* Ensure to show more information of an error

issue #1425, #1482


cosmo0920:
I've confirmed following things:

* rabbitmq user does not have adequate permissions for vhost and when error occurred
  * Try to reconnect
  * Exit normally with Keyboard interrupt & SIGHUP signal

ashie:
Thanks!

PR #1490, opened 2015-07-23
[AMQPConnection] Lengthen heartbeat check to receive larger data

ashie:
1 [sec] -> 60 [sec]

issue #1425


cosmo0920:
I've confirmed that disconnection is no longer occurred when receiving large data.

ashie:
Thanks!

PR #1493, opened 2015-07-24
[EventsView] Filter events by EventType instead of TriggerStatus

ashie:
EventType should be used to filter events instead of TriggerStatus because the "Status" column
of EventsView uses EventType, not TriggerStatus.

Issue #1491


ashie:
Thanks!

PR #1494, opened 2015-07-26
Log status periodically (fluentd).

kz0817:
The current periodic logging feature doesn't works for fluentd plugin,
because Hap2Fluentd doesn't use BasePoller.

This patch makes it possible.

### DIFF
--- a/server/hap2/hatohol/hap2_fluentd.py
+++ b/server/hap2/hatohol/hap2_fluentd.py
@@ -25,6 +25,7 @@ import datetime
 import json
 import re
 import time
+import signal
 from hatohol import haplib
 from hatohol import standardhap
 
@@ -50,6 +51,7 @@ class Hap2FluentdMain(haplib.BaseMainPlugin):
         self.__launch_args = args.fluentd_launch.split(" ")
         self.__accept_tag_reg = args.tag
         self.__accept_tag_pattern = re.compile(self.__accept_tag_reg)
+        self.__log_status_interverl = args.log_status_interval
 
     def set_ms_info(self, ms_info):
         self.__ms_info = ms_info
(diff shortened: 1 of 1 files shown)


server/hap2/hatohol/hap2_fluentd.py
```diff
@@ -50,6 +51,7 @@ def set_arguments(self, args):
         self.__launch_args = args.fluentd_launch.split(" ")
         self.__accept_tag_reg = args.tag
         self.__accept_tag_pattern = re.compile(self.__accept_tag_reg)
+        self.__log_status_interverl = args.log_status_interval
```
cosmo0920: Typo?

```diff
-interverl
+interval
```

server/hap2/hatohol/hap2_fluentd.py
```diff
@@ -135,7 +149,8 @@ def __parse_line(self, line):
             timestamp, tag = self.__parse_header(header)
         except:
             timestamp, tag, msg = None, None, None
-            haplib.handle_exception()
+            exctype, value = haplib.handle_exception()
```
cosmo0920: This `exctype` is no longer used?

If so, we should use `_` instead of `exctype` to clarify "no longer used".

ashie:
Several tests fail:

https://travis-ci.org/project-hatohol/hatohol/jobs/72658373

server/hap2/hatohol/hap2_fluentd.py
cosmo0920: It should be fix Hap2Fluentd in test. Because for now, it does not specify `arg.log_status_interval`: https://github.com/project-hatohol/hatohol/blob/b2ea925a6d8c78ba5d6dfcd723576f73c894fb63/server/hap2/hatohol/test/TestHap2Fluentd.py#L176-L185

kz0817:
Thank you for the review and I'm sorry for slow response.
I'll check it.

kz0817:
I sent the revised version as #1506.
So I close this PR.

PR #1497, opened 2015-07-28
[hatohol.spec] Add "--with-qpid" to configure script

ashie:
It's disabled by default, but the RPM package still includes
hatohol-arm-plugin-zabbix which depends on qpid.

PR #1498, opened 2015-07-28
[hatohol.spec] Add "--with-qpid" to configure script

ashie:
It's disabled by default, but the RPM package still includes
hatohol-arm-plugin-zabbix which depends on Qpid.


ashie:
It's completely same with #1497.

Although building on Werker will be failed, it's not this patch's problem.
We should modify Dockerfile to solve it.

PR #1508, opened 2015-08-05
[EventsView] Initial implementation of the time range filter

ashie:
We are planning to add a time range filter for events to the next version.
So I made a prototype for it.
The page layout will be modified later (until the next release).


kz0817:
I think it's better to add a way to clear the begin time and the end time quickly.

PR #1511, opened 2015-08-09
travis: use `sudo: required` explicitly

cosmo0920:
Because current build steps in Travis is heavily depends on sudo.

Upgrading document is here: http://docs.travis-ci.com/user/migrating-from-legacy/
But, sadly, I have no idea to upgrade Travis build step to adopt new Travis infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 notifications:
   email:
     recipients:


kz0817:
With this patch, can any 'sudo' be removed in .travis.yaml ?
Or do we have to do so due to TravsCI 's spec.?

cosmo0920:
> With this patch, can any 'sudo' be removed in .travis.yaml ?
> Or do we have to do so due to TravsCI 's spec.?

Ah, we can remove 'sudo' with apt addons like this:

```yaml
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test # for ppa
    packages:
    - cmake # for apt-get
```

And we should use ppa repository instead of using setup script. (for cutter)

kz0817:
Can we remove 'sudo' in the current .travis.yaml ?
For example, the head in the following line.

>  - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y

cosmo0920:
>> Can we remove 'sudo' in the current .travis.yaml ?
>> For example, the head in the following line.
> 
>  *       sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y

For new Travis Infrastructure, above line becomes like this:

```yaml
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
```

But I have no idea how to treat following lines in new Travis Infrastructure:

```yaml
   - sudo /etc/init.d/qpidd restart
   - sudo server/misc/setup-rabbitmq-server-port.sh
   - sudo service rabbitmq-server restart
```

kz0817:
Oh. I got it. 

'sudo: required' is effective for 'apt:', doesn't it?

Issue #1446, opened 2015-07-13
Some test materials in server/test/DBTablesTest.cc should be immutable


kz0817:
Some test materials such as testServerTypeInfo and testServerStatus
aren't defined without 'const'. This situation may cause unexpected change of them
by mistake and wrong test results.

kz0817:
This issue was addressed.

PR #1513, opened 2015-08-11
Immutable built in test material #1446

kz0817:
These patches makes built-in test materials immutable in order to
prevent them from being changed by mistake.

### DIFF
--- a/server/test/DBTablesTest.cc
+++ b/server/test/DBTablesTest.cc
@@ -1054,7 +1054,7 @@ const IncidentTrackerInfo testIncidentTrackerInfo[] = {
 	"o.o662L6q1V7E",          // password
 }
 };
-size_t NumTestIncidentTrackerInfo = ARRAY_SIZE(testIncidentTrackerInfo);
+const size_t NumTestIncidentTrackerInfo = ARRAY_SIZE(testIncidentTrackerInfo);
 
 IncidentInfo testIncidentInfo[] = {
 {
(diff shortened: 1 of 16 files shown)


cosmo0920:
It seems to be good.

But I have a question.
C++11 reference uses following style range based for:

```cpp
attr(optional) for ( range_declaration : range_expression ) loop_statement		
``` 

Why do you use `range_declaration: range_expression` style?

kz0817:
> Why do you use range_declaration: range_expression style?

I wasn't aware that.
What is 'C++11 reference' you mentioned ?
Could you please tell me the pointer to that ?

cosmo0920:
Oh, sorry.

This is cpp reference: http://en.cppreference.com/w/cpp/language/range-for

kz0817:
> Here is cpp reference: http://en.cppreference.com/w/cpp/language/range-for#Syntax

OK. I'll fix to follow the style.

kz0817:
I fixed. Please review again.

cosmo0920:
:+1:

server/test/DBTablesTest.cc
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013-2014 Project Hatohol
+ * Copyright (C) 2013-2015 Project Hatohol
```
cosmo0920: If this change including in this PR, please update copy right years in other headers and sources.

kz0817: OK. I'll modify other source files.

cosmo0920: Thanks!

kz0817: I've done it. Please review again.

cosmo0920:
@ashie Any objections? If nothing, I'll merge this PR.

ashie:
No problem.

kz0817:
おーきに。

PR #1514, opened 2015-08-14
[hap2] Abort launching hatohol plugin if the PID file couldn't be created.

kz0817:
Currently the failure of the creation of the PID file is ignored.
However, the execution should be terminated on such situation.


kz0817:
I noticed this is imcopleted. I once close the PR.

Issue #1461, opened 2015-07-13
hap2: Dynamic detection of transporter modules.


kz0817:
Plugins assumes python modules under hatohol/transporters as transporter modules.
They will try to search such file at start up and list up transporter modules automatically.

Target date: 7/27 (Mon)

kz0817:
The feature was merged. I close this issue.

PR #1516, opened 2015-08-14
[hap2] find and load transporter modules dynamically (#1461)

kz0817:
In HAPI2.0 framework, a transporter module is replaceable.
To improve the flexibility, this patch tries to find transporter
modules: hatohol/transporters under each $PYTHONPATH.

### DIFF
--- a/server/hap2/hatohol/transporters/__init__.py
+++ b/server/hap2/hatohol/transporters/__init__.py
@@ -1,0 +1,27 @@
+import os
+from hatohol import transporter
+
+"""
+Load all pyton files in this directory except for files which begin
+with hyphen.
+"""
+
+def is_valid_module_name(filename):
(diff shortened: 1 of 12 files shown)


server/hap2/hatohol/transporters/__init__.py
```diff
@@ -0,0 +1,27 @@
+import os
+from hatohol import transporter
+
+"""
+Load all pyton files in this directory except for files which begin
+with hyphen.
+"""
+
+def is_valid_module_name(filename):
+    if filename[-3:] != ".py":
```
ashie: BTW I prefer using "startswith" and "endswith" methods.

kz0817: Thank you for the good suggestion.
I'll use it.

PR #1520, opened 2015-08-17
Retry when eintr happens #1519

kz0817:
Some hap2 tests failed with IOError due to EINTR such as
https://travis-ci.org/project-hatohol/hatohol/jobs/75575033.

This patch provides a wrapper object that retries in it.

### DIFF
--- a/server/hap2/hatohol/hap.py
+++ b/server/hap2/hatohol/hap.py
@@ -66,7 +66,7 @@ class Signal:
         self.restart = restart
         self.critical = critical
 
-def MultiprocessingQueue():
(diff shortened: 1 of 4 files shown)


kz0817:
I turn down this PR because the add code has the following problem.
I will change the structure to make it possible to test it.

https://travis-ci.org/project-hatohol/hatohol/jobs/75915569
<pre>
    if self.__reply_queue.get(True, self.__timeout_sec) == True:

  File "/home/travis/build/project-hatohol/hatohol/server/hap2/hatohol/hap.py", line 82, in __get

    if errno != errno.EINTR:

AttributeError: 'int' object has no attribute 'EINTR'
</pre>

PR #1531, opened 2015-08-20
Suppress warning of host cache #1424 2

kz0817:
This pull request suppress warning of HostInfoCache.
The warning is shown when the correspond host is not found in the DB.
However, some HAPs such as hap2_fluentd.py doesn't send host information.
In that case, the warning is shown repeatedly. The patches automatically
unknown hosts register in HostInfoCache ad hoc.

### DIFF
--- a/server/src/HostInfoCache.cc
+++ b/server/src/HostInfoCache.cc
@@ -132,6 +132,7 @@ void HostInfoCache::registerAdHoc(const LocalHostIdType &idInServer,
                                   const string &name, Element &elem)
 {
 	ServerHostDef svHostDef;
+	svHostDef.id             = -1;
 	svHostDef.hostId         = INVALID_HOST_ID;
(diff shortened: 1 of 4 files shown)


server/src/HostInfoCache.cc
```diff
@@ -128,6 +128,21 @@ bool HostInfoCache::getName(
 	return found;
 }
 
+void HostInfoCache::registerAdHoc(const LocalHostIdType &idInServer,
+                                  const string &name, Element &elem)
+{
+	ServerHostDef svHostDef;
```
ashie: svHostDef.id isn't initialized.
Although it's not used in the current implementation, it may cause problems in the future.

kz0817: OK. I initialize it.

kz0817:
I fixed the leak of the initialization and ran the test on my local machine.
Please review again.

kz0817:
10Q

PR #1535, opened 2015-08-27
Implement trigger updated option2

Mnakagawa:
Issue #1452

### DIFF
--- a/server/hap2/hatohol/hap2_nagios_ndoutils.py
+++ b/server/hap2/hatohol/hap2_nagios_ndoutils.py
@@ -119,7 +119,7 @@ class Common:
         return server, port, database
 
     def __convert_to_nagios_time(self, hatohol_time):
-        over_second= hatohol_time.split(".")[0]
+        over_second = hatohol_time.split(".")[0]
         time_array = [char for char in over_second]
         time_array.insert(4, "-")
         time_array.insert(7, "-")
@@ -194,8 +194,11 @@ class Common:
             in_cond = "','".join(host_ids)
             sql += " WHERE %s.host_object_id in ('%s')" % (t2, in_cond)
 
+        def all_triggers_should_send():
+            return fetch_id is None
+
         update_type = "ALL"
-        if fetch_id is None:
+        if all_triggers_should_send():
             if self.__trigger_last_info is None:
                 self.__trigger_last_info = self.get_last_info("trigger")
 
--- a/server/hap2/hatohol/hap2_nagios_ndoutils.py
+++ b/server/hap2/hatohol/hap2_nagios_ndoutils.py
@@ -117,6 +117,16 @@ class Common:
 
         return server, port, database
 
+    def __convert_to_nagios_time(self, hatohol_time):
+        over_second= hatohol_time.split(".")[0]
+        time_array = [char for char in over_second]
+        time_array.insert(4, "-")
+        time_array.insert(7, "-")
+        time_array.insert(10, " ")
+        time_array.insert(13, ":")
+        time_array.insert(16, ":")
+        return "".join(time_array)
+
     def collect_hosts_and_put(self):
         sql = "SELECT host_object_id,display_name FROM nagios_hosts"
         self.__cursor.execute(sql)
--- a/server/hap2/hatohol/hap2_nagios_ndoutils.py
+++ b/server/hap2/hatohol/hap2_nagios_ndoutils.py
@@ -59,6 +59,7 @@ class Common:
         self.__db_user = self.DEFAULT_USER
         self.__db_passwd = ""
         self.__time_offset = datetime.timedelta(seconds=time.timezone)
+        self.__trigger_last_info = None
 
     def close_connection(self):
         if self.__cursor is not None:
@@ -125,6 +126,7 @@ class Common:
         time_array.insert(10, " ")
         time_array.insert(13, ":")
         time_array.insert(16, ":")
+
         return "".join(time_array)
 
     def collect_hosts_and_put(self):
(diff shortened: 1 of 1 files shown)


server/hap2/hatohol/hap2_nagios_ndoutils.py
```diff
@@ -117,6 +118,17 @@ def __parse_url(self, url):
 
         return server, port, database
 
+    def __convert_to_nagios_time(self, hatohol_time):
+        over_second= hatohol_time.split(".")[0]
```
kz0817: missing a space!
over_second= ->
over_second =

server/hap2/hatohol/hap2_nagios_ndoutils.py
```diff
@@ -117,6 +118,17 @@ def __parse_url(self, url):
...
+        time_array = [char for char in over_second]
+        time_array.insert(4, "-")
+        time_array.insert(7, "-")
+        time_array.insert(10, " ")
+        time_array.insert(13, ":")
+        time_array.insert(16, ":")
+
```
kz0817: I seems too lower operations. That degrades the readability and maintainability.
How about making a function that convert a hapi_time to a datetime object in haplib and using strftime().

server/hap2/hatohol/hap2_nagios_ndoutils.py
```diff
@@ -182,12 +194,16 @@ def collect_triggers_and_put(self, fetch_id=None, host_ids=None):
...
             sql += " WHERE %s.host_object_id in ('%s')" % (t2, in_cond)
 
-        # NOTE: The update time in the output is renewed every status
-        #       check in Nagios even if the value is not changed.
-        # TODO by 15.09:
-        #   We should has the previous result and compare it here in order to
-        #   improve performance. Or other columns such as last_state_change in
-        #   nagios_servicestatus might be used.
+        update_type = "ALL"
+        if fetch_id is None:
```
kz0817: The purpose of this condition should be indicated in order to improve readability.

For exampe.

def all_triggers_should_send():
    return fetch_id is None

if all_triggers_should_send():
   ....

server/hap2/hatohol/hap2_nagios_ndoutils.py
Mnakagawa: >  it looks this function is only used as making a last_info. 

That's wrong.
It used to create SQL sentence. 
datetime object can not use sql.
So, we should use str() as last_info.

server/hap2/hatohol/hap2_nagios_ndoutils.py
Mnakagawa: sure

server/hap2/hatohol/hap2_nagios_ndoutils.py
Mnakagawa: k

server/hap2/hatohol/hap2_nagios_ndoutils.py
Mnakagawa: The following

https://github.com/project-hatohol/hatohol/blob/implement-trigger-updated-option2/server/hap2/hatohol/hap2_nagios_ndoutils.py#L204

Mnakagawa: I use Timestamp type of HAPI2.0 as a trigger last_info.

Mnakagawa: Issue #1452 deadline is today.
I want to any response.

kz0817: Making an array of character is uncool and may degrade performance.

Mnakagawa: I think so too.
But, python does not have a way to insert a charactor to string.
If you know good idea, please tell me.

kz0817: One thing is to use a datetime object converted from hatohol_time.

Other is to concatenate partial strings of hatohol time such as
x = hatohol_time[0:4] + "-" + hatohol_time[4:6] + "-".
(However, raw numbers are ugly)

server/hap2/hatohol/hap2_nagios_ndoutils.py
```diff
@@ -182,12 +194,19 @@ def collect_triggers_and_put(self, fetch_id=None, host_ids=None):
             in_cond = "','".join(host_ids)
             sql += " WHERE %s.host_object_id in ('%s')" % (t2, in_cond)
 
-        # NOTE: The update time in the output is renewed every status
-        #       check in Nagios even if the value is not changed.
-        # TODO by 15.09:
-        #   We should has the previous result and compare it here in order to
-        #   improve performance. Or other columns such as last_state_change in
-        #   nagios_servicestatus might be used.
+        def all_triggers_should_send():
```
kz0817: This tiny function may be simpler when we use lambda.
<pre>
all_triggers_should_send = lambda: fetch_id is None
</pre>

server/hap2/hatohol/hap2_nagios_ndoutils.py
Mnakagawa: > One thing is to use a datetime object converted from hatohol_time.

You mean the following?
hatohol_time(string) -> datatime object (using strptime) -> nagios_time(string. using strftime)

kz0817: YES.

Mnakagawa: I understand.
It good idea!

I close this PR and will resend new PR.

PR #1537, opened 2015-08-31
[HostResourceQueryOption] Filter by plural {servers,hostgroups,hosts}

ashie:
Add member functions to set filters by plural {servers,hostgroups,hosts}.


ashie:
setTarget{Server,Hostgroup,Host}Id() are still remained to keep compatibility.

ashie:
Sorry, I'll send another pull request to improve (enable to set both select & exclude filters).

PR #1538, opened 2015-08-31
[HostResourceQueryOption] Filter by plural {servers,hostgroups,hosts}

ashie:
Add member functions to set filters by plural {servers,hostgroups,hosts}.

It's the revised version of #1537.

### DIFF
--- a/server/src/HostResourceQueryOption.cc
+++ b/server/src/HostResourceQueryOption.cc
@@ -66,6 +66,7 @@ struct HostResourceQueryOption::Impl {
 	LocalHostIdType targetHostId;
 	HostgroupIdType targetHostgroupId;
 	bool            excludeDefunctServers;
+	list<ServerIdType> excludeServerIdList;
 
 	Impl(const Synapse &_synapse)
 	: synapse(_synapse),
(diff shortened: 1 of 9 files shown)


server/src/HostResourceQueryOption.cc
```diff
@@ -436,19 +548,227 @@ string HostResourceQueryOption::makeConditionForNormalUser(
...
+	string serverIdColumnName = getServerIdColumnName();
+	string hostgroupIdColumnName = getHostgroupIdColumnName();
+
+	for (auto &pair: m_impl->selectedServerHostgroupSetMap) {
+		if (!isAllowedServer(pair.first))
+			continue;
+		for (auto &hostgroupId: pair.second) {
+			if (!isAllowedHostgroup(pair.first, hostgroupId))
+				continue;
+			addCondition(condition,
```
kz0817: This is just an impression. No need to take care of it.

We want to define a new class that handles a condition string, provides utility methods to concatenate addition conditions and returns C style string pointer.

ashie:
Thanks!

PR #1540, opened 2015-09-04
Can add date and time to the package version.

kz0817:
Developers and tester sometimes would like to use a temporarily
built package. However, it's needed to have a large version number
update the package on the test platfrom.

This patch enables to include date and time to the version string
automatically.


kz0817:
An example of the package name with date and time:
hatohol-server-15.07_dev1_20150904_062618-1.el7.centos.x86_64.rpm

kz0817:
Oh! build failed. I turn down the PR once.

PR #1542, opened 2015-09-04
[server] Add a simple IncidentSender "IncidentSenderHatohol"

ashie:
Add a simple IncidentSender "IncidentSenderHatohol" which manages incidents without external
programs. It generates incidents by itself and stores them to Hatohol's DB directly.

TODO:

* Add REST API to update incidents
* Modify WebUI to support it


ashie:
Thanks!

PR #1544, opened 2015-09-07
[server] Add REST API for incident

ashie:
This modification adds REST API for incident.
Only PUT method (updating an incident) is allowed.


ashie:
Thanks!

PR #1550, opened 2015-09-14
[server] Add plural severities and statuses filter for events

ashie:
Example queries for them:

* /event?severities=2%2C4
* /event?statuses=0%2C2

Decodeded query parts:

* severities=2,4 (WARNING & CRITICAL)
* statuses=0,2 (OK & UNKNOWN)


ashie:
Thanks!

PR #1553, opened 2015-09-14
[client][EventsView] Add a mechanism to customize visible columns on the events table

ashie:
The UI for customizing visible columns isn't implemented yet.
I'll add it later.

In addition, I'll also modify the default order of columns later.


ashie:
Thanks!

PR #1556, opened 2015-09-15
[server][FaceRest::ResourceHandler] Eleminate redundant lines.

kz0817:
Some sub classes of ResoureHandler had to implement similar manadatory
methods such as handler(). This patch addressed the boring work by
calling the handler in the base class: ResourceHandler instead of
each sub class.


kz0817:
Good morning and Thank you.

2015年9月16日水曜日、Takuro Ashie<notifications@github.com>さんは書きました:

PR #1565, opened 2015-09-21
[client] Add a config dialog for EventsView

ashie:
Enable to customize following items:

* auto reload interval
* number of rows
* visible columns
    * Note that incident columns aren't show when no incident setting exists.

In addition I removed stupidtable dependency of EventsView because it's not used anymore.

events-view-config


client/static/css/hatohol.css
```diff
@@ -73,6 +73,34 @@ div.dialog {
...
+#events-view-config ul.nav-tabs {
+    width: 100%;
+}
+
+#events-view-config .tab-content {
+    padding: 16px 4px 16px 4px;
+}
+
+#events-view-config input[type="text"] {
+    width:4em;
```
kz0817: a space just after colon is missing.

ashie: Thanks. I added it.

ashie:
Thanks!

PR #1566, opened 2015-09-24
[server] Add REST API to get only number of events

ashie:
An example query:

* /events?countOnly=true&serverId=3 (Any event filters can be added)
    
An example response:

    {
        "apiVersion": 4,
        "errorCode": 0,
        "numberOfEvents:4"
    }

### DIFF
--- a/server/src/DBTablesMonitoring.cc
+++ b/server/src/DBTablesMonitoring.cc
@@ -2540,6 +2540,23 @@ HatoholError DBTablesMonitoring::getNumberOfMonitoredItemsPerSecond
 	return HatoholError(HTERR_OK);
 }
 
+size_t DBTablesMonitoring::getNumberOfEvents(const EventsQueryOption &option)
+{
+	DBClientJoinBuilder builder(tableProfileEvents, &option);
+	string stmt =
+	  StringUtils::sprintf("count(distinct %s)",
+	    option.getColumnName(IDX_EVENTS_UNIFIED_ID).c_str());
+	DBAgent::SelectExArg &arg = builder.build();
+	arg.add(stmt, SQL_COLUMN_TYPE_INT);
+
+	getDBAgent().runTransaction(arg);
+
+	// get the result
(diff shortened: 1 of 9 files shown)


server/src/DBTablesMonitoring.cc
```diff
@@ -2540,6 +2540,23 @@ HatoholError DBTablesMonitoring::getNumberOfMonitoredItemsPerSecond
 	return HatoholError(HTERR_OK);
 }
 
+size_t DBTablesMonitoring::getNumberOfEvents(const EventsQueryOption &option)
+{
+	DBClientJoinBuilder builder(tableProfileEvents, &option);
+	string stmt =
```
kz0817: Just impression: Don't take care.
We want use 'constexpr' for this kind of lines.
As a result, SQL statements are generated on the build time, instead of runtime.

server/src/RestResourceHost.cc
```diff
@@ -487,6 +488,13 @@ HatoholError RestResourceHost::parseEventParameter(EventsQueryOption &option,
 		option.setTriggerStatuses(statuses);
 	}
 
+	// countOnly
+	value = static_cast<const gchar*>(
+	  g_hash_table_lookup(query, "countOnly"));
+	if (value && *value && string(value) == "true") {
```
kz0817: How about returning an error or logging that when the RHS of countOnly is neither "true" or false ?

ashie: I'll add an error.

server/src/DBTablesMonitoring.cc
cosmo0920: Hmm..., if we make this value to `constexpr`, its type should be `char` instead of `std::string`.

kz0817: Probably, we have to do so.

ashie: In this case the column name is determined at runtime, so stmt is also built at runtime.
I don't modify it.

ashie:
Thanks!

PR #1570, opened 2015-09-25
[client] Add API to store event filters

ashie:
It's almost same with graph settings.
So I use a same method to implement them.


ashie:
Thanks!

PR #1571, opened 2015-09-25
[client][EventsView] Colorize all columns

ashie:
Default colors will be changed later.

colorize-all-columns


ashie:
Sorry, some tests are failed.
I'll fix them.

PR #1573, opened 2015-09-25
[client][EventsView] Colorize all columns

ashie:
Default colors will be changed later.

colorize-all-columns

It's the revised version of #1571


ashie:
Thanks!

PR #1575, opened 2015-09-28
[server] Can initialize tables for Hatohol on another machine.

kz0817:
Previously hatohol-db-initiator is hard to initialize tables in
the DB server on an another machine.

This patch provides the way to specify the DB server and the port.

In addtion, some paramters for hatohol-db-initiator are set
by default to make it easy to use.


kz0817:
Oops! Some tests failed due to the change of command line option.
For example,

--db_name ->
--db-name

I close this issue, and then I'll submit PR after the fix.

kz0817:
I fixed the problem and reopen this issue.
Please review the PR.

cosmo0920:
:+1:

PR #1583, opened 2015-09-30
[client][EventsView] Add UI for creating filter definitions (WIP)

ashie:
I added UI to define filters.
It's not completed yet.

* "Default filter" setting isn't implemented yet
* No one uses created filters yet.

It's just only create filter settings (JSON) and store them to DB.

Note that this page isn't shown by default.
You can see it by the following query:

http://localhost/ajax_events?devel=true

filter-settings

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -121,13 +121,17 @@ div#filter-name-config {
 }
 
 input#filter-name-entry {
-    width: 420px;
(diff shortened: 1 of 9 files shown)


cosmo0920:
Oh, this PR seems to be conflicted....
Could you solve conflict?

ashie:
#1587

PR #1587, opened 2015-10-01
[client][EventsView] Add UI for creating filter definitions (WIP)

ashie:
Revice #1583 

I added UI to define filters.
It's not completed yet.

* "Default filter" setting isn't implemented yet
* No one uses created filters yet.

It's just only create filter settings (JSON) and store them to DB.

Note that this page isn't shown by default.
You can see it by the following query:

http://localhost/ajax_events?devel=true

filter-settings

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -121,13 +121,17 @@ div#filter-name-config {
 }
 
 input#filter-name-entry {
-    width: 420px;
+    width: 380px;
     margin-right: -6px;
     display: inline-block;
 }
 
 button#filter-name-list-button {
     margin-bottom: 2px;
+}
+
+button#filter-name-list-delete-button {
+    margin-bottom: 6px;
 }
 
 input#term-filter-setting {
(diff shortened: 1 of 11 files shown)


kz0817:
A test failed. Can we ignore it ?
<pre>
  1) EventsViewConfig "before each" hook:

     Uncaught Error: expected undefined to equal 'none' (http://localhost:8000/test/expect.js:102)
</pre>

kz0817:
I built this branch and ran. Different from the screen you attached in this PR, a tab was not shown on the dialog as the following picture. Is this correct ?
event-consle-setting

ashie:
> A test failed. Can we ignore it ?

No. I'll investigate it.

ashie:
> I built this branch and ran. Different from the screen you attached in this PR

Could you add "?devel=true" to the URL?

> Note that this page isn't shown by default.
> You can see it by the following query:
> 
> http://localhost/ajax_events?devel=true

kz0817:
> Could you add "?devel=true" to the URL?

Oh. I missed it. OK. I confirmed that a tab is shown.

ashie:
> A test failed. Can we ignore it ?

I fixed it (although it's an issue in #1584).
Please review again.

ashie:
Thanks!

PR #1593, opened 2015-10-04
[client] Add an ability to change incident statuses

ashie:
Note that this feature is enabled only when the incident tracker type "Hatohol" is enabled
and "Redmine" is disabled. Because incident status values are different between both types,
it's hard to enable it against both types by the unified menu.

change-treatment


cosmo0920:
:+1:

PR #1596, opened 2015-10-07
[EventsView] Hide toggle-sidebar button by default

ashie:


ashie:
Merge it by myself because it's a trivial fix.

PR #1598, opened 2015-10-07
[WebUI] Make DEBUG False to avoid internal information from being shown.

kz0817:
When the DEBUG variable is True and any problem happens in WebUI code,
the debug information is sent to the viewer. This behavior is generally
undesired.

### DIFF
--- a/client/HowToTest.md
+++ b/client/HowToTest.md
@@ -62,11 +62,11 @@ Ex.)
 
 run djang with debug mode (set an environment variable: HATOHOL_DEBUG=1)
 
-    $ HATOHOL_DEBUG=1 ./manage.py runserver 0.0.0.0:8000
+    $ HATOHOL_DEBUG=1 ./manage.py runserver --insecure 0.0.0.0:8000
 
 or if you want to use a different database on a test, you can specify the test settings that use the database named 'test_hatohol_client'.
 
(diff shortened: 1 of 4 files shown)


ashie:
Client tests are failed with the following message:

```
Error loading resource http://localhost:8000/static/js.external/jquery.js (203). Details: Error downloading http://localhost:8000/static/js.external/jquery.js - server replied: NOT FOUND
...
```

kz0817:
Oh! I will check it.

kz0817:
According to the following Django manual, the static files can be served only when the debug mode.

https://docs.djangoproject.com/en/1.5/howto/static-files/ 

> This helper function works only in debug mode and only if the given prefix is local (e.g. /static/) and not a URL (e.g. http://static.example.com/).

The following page describes about '--insecure' option that is an easy way to work around this. 
https://docs.djangoproject.com/en/1.5/ref/contrib/staticfiles/

This option is insecure as its name, however, we can use it only for the test.

ashie:
> Add --insecure option to serve static files

I think it's not good idea.
I want to use DEBUG=True for development.

I propose to use different settings for development server & production server.

* Add hatohol/settings_development.py & hatohol/settings_production.py
* Load settings_development.py at hatohol/wsgi.py
* Load settings_production.py at conf/apache/django.wsgi
* Common settings can be imported by "from settings import *"

kz0817:
I see. I agree on your suggestion and I'll modify that part according to the idea

kz0817:
I once close this issue.

PR #1599, opened 2015-10-08
[server] Filter events by incident statuses

ashie:
Example query:

* /event?incidentStatuses=NONE%2CHOLD

Decodeded query part:

* incidentStatuses=NONE,HOLD

I resolved issues in #1595.


ashie:
Thanks!

PR #1600, opened 2015-10-08
[EventsView] Calculate events table's height more strictly

ashie:


kz0817:
Some tests failed. But It seems that they are not related with the change. What do you think ?

ashie:
Yes, they aren't related with this.

cosmo0920:
Looks good for me.

ashie:
Thanks!

PR #1604, opened 2015-10-09
Brush up production web ui config (#1602)

kz0817:
These patches enables the path of WebUI to be any path.


kz0817:
Oh! The test has broken.
I'll fix it.

kz0817:
I submitted an another PR (#1606) and close this issue.

PR #1606, opened 2015-10-12
Brush up production web ui config 2 (#1602)

kz0817:
These patches enables users to set any WebUI path.
This is a revised version of #1604


kz0817:
I turn down this pull request once to do rebase for removing the patch 11b215e that has already been merged.

PR #1609, opened 2015-10-13
Implement important rate privileges

cosmo0920:
* Add severity rank settings UI into WebUI
* Add severity rank privileges
  * constants
  * constraint to get severity ranks user defined settings
* Add migration for administrator user privileges


cosmo0920:
Thanks!

PR #1610, opened 2015-10-13
[EventsView] Add clear buttons to time range filter entries

ashie:
time-range-filter-clear-button


cosmo0920:
Looks good to me!

PR #1612, opened 2015-10-14
[EventsView] Add "Treatment" (a.k.a incident status) filter

ashie:
treatment-filter

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -54,6 +54,11 @@ div.dialog {
   display: none;
 }
 
+div#controller-container
+{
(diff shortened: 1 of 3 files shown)


ashie:
In addition, some minor fixes are included:

* Swap the position of "Status" filter & "Minimum Severity" filter to follow the default order of events table's columns
* Avoid shrinking events table's width when the incident feature isn't enabled

cosmo0920:
Umm..., merging #1610 causes conflicts.... :sweat: 
Could you solve conflicts?

ashie:
I resolved it in #1616.

PR #1616, opened 2015-10-15
[EventsView] Add "Treatment" (a.k.a incident status) filter

ashie:
treatment-filter

In addition, some minor fixes are also included in thie PR:

* Swap the position of "Status" filter & "Minimum Severity" filter to follow the default order of events table's columns
* Avoid shrinking events table's width when the incident feature isn't enabled

This is the rebased version of #1612.

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -54,6 +54,11 @@ div.dialog {
   display: none;
 }
 
+div#controller-container
+{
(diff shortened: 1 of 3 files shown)


cosmo0920:
Works well in my environment.
After CIs finished, I'll merge this PR.

ashie:
Thanks!

PR #1617, opened 2015-10-15
[Wercker] Try to use CentOS 6.7 docker image

cosmo0920:
Wercker says: age (6.6) from centos, Server error: 400 trying to fetch remote history.

It seems that centos:6.6 image has been deleted from DockerHub.
We should start using CentOS 6.7 image in Wercker.


cosmo0920:
Thanks!

PR #1619, opened 2015-10-16
[REST API][SeverityRank] Severity rank REST API fix

cosmo0920:
I've found missing label and as_important in severity_ranks table and its REST API.... :'(

Follow up #1618.

### DIFF
--- a/server/src/DBTablesConfig.cc
+++ b/server/src/DBTablesConfig.cc
@@ -1656,6 +1656,15 @@ void DBTablesConfig::getIncidentTrackerIdSet(
 	}
 }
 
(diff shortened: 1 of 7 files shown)


server/src/DBTablesConfig.cc
```diff
@@ -1892,6 +1960,15 @@ string SeverityRankQueryOption::getCondition(void) const
 		addCondition(condition, colorCondition);
 	}
 
+	if (!m_impl->label.empty()) {
+		DBTermCStringProvider rhs(*getDBTermCodec());
+		string colorCondition = StringUtils::sprintf(
```
ashie: どこがcolorConditionやねん...

PR #1621, opened 2015-10-17
[Wercker] Rewrite wercker.yml to adopt new Docker based building stack

cosmo0920:
Before merging this PR, it also needs to switch Wercker build settings from classic.

Related to #1615.

Yes, new Wercker build stack can use Docker image directly!!

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -1,9 +1,63 @@
-box: wercker-labs/docker
+box: centos:6.7
 build:
   steps:
     - script:
-        name: Build a CentOS 6.7 Docker Hatohol image box
+        name: Install libraries for Hatohol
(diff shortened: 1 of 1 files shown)


wercker.yml
```diff
@@ -1,9 +1,63 @@
-box: wercker-labs/docker
+box: centos:6.7
```
cosmo0920: In new Wercker build stack, this field should be used for specifying DockerHub's image name.

cosmo0920:
@ashie I've already configured to use new Wercker build stack which is based on Docker. Could you review this PR?

cosmo0920:
Thanks!

PR #1623, opened 2015-10-18
[EventsView] Add default filter config & filter chooser

ashie:
Default filter:

default-filter

Filter chooser:

filter-chooser

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -166,6 +166,42 @@ input#quick-host-search-entry {
     display: inline-block;
 }
 
+div.view-layout-mock {
+    display: inline-block;
+    vertical-align: middle;
+}
+
+table.view-layout-mock {
+    border-width: 2px;
+    border-color: #aaaaaa;
+    width: 150px;
+    height:120px;
+}
+
+table.view-layout-mock td.sidebar {
+    border-width: 2px;
+    border-color: #aaaaaa;
+    width: 50px;
+}
+
+table.view-layout-mock td.controlbar {
+    border-width: 2px;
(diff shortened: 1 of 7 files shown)


ashie:
Now they are enabled by default.
You don't need "?devel=true" to see it any more.

kz0817:
I met the following error. Is this due to the update from a development version to another development version ?
<pre>
$ PATH=hap/.libs:$PATH LD_LIBRARY_PATH=mlpl/src/.libs:src/.libs:common/.libs:hap/.libs src/.libs/hatohol --pid-file ~/tmp/hatohol.pid --foreground
[INFO] <main.cc:217> started hatohol server: ver. 15.07_dev1
[INFO] <DBTables.cc:203> DBTables 16 is outdated, try to update it ...
        old-version: 00.16 [00]
        new-version: 00.18 [00]
[CRIT] <main.cc:175> Failed to create a database object. This program is aborted. Reason: <DBAgentMySQL.cc:667> Failed to query: ALTER TABLE severity_ranks ADD COLUMN label VARCHAR(255) NOT NULL, ADD COLUMN as_important INT(11) NOT NULL DEFAULT 1: (1060) Duplicate column name 'label'
</pre>

ashie:
> Is this due to the update from a development version to another development version ?

Probably, yes it is.
I also upgraded  the DB using this code, but it's successfully completed.

But we should check it again. I'll do it later.

kz0817:
Once I dropped the database 'hatohol' and did hatohol-db-initiator, this branch of hatohol ran successfully. Thank for the comment.

ashie:
Thanks!

PR #1624, opened 2015-10-19
[EvnetsView] Add tooltips for event descriptions

ashie:


ashie:
Oh, sorry. Some tests fail.
I'll fix them.

PR #1625, opened 2015-10-19
[HatoholDBInitiator] Add severity-ranks target and sql file

cosmo0920:
Before implementing statistics, we should have a severity_ranks fixture into database.


cosmo0920:
Thanks!

PR #1632, opened 2015-10-21
Dont read config file on conf manager test

kz0817:
In hatoholInit(), ConfigManager::reset() is invoked with
loadConfigFile = true. This makes the test results depned on
contents of the configuration file.
Here, we call reset() with loadConfigFile = false in order to
eliminate the dependency and makes sure the same conditions.

### DIFF
--- a/server/test/testConfigManager.cc
+++ b/server/test/testConfigManager.cc
@@ -284,8 +284,7 @@ void test_parseFaceRestNumWorkers(gconstpointer data)
 	cmds.activate();
 
 	cppcut_assert_equal(
-		ConfigManager::getInstance()->getFaceRestNumWorkers(),
-		expect);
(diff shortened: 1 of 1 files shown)


server/test/testConfigManager.cc
```diff
@@ -33,6 +33,14 @@ namespace testConfigManager {
 void cut_setup(void)
 {
 	hatoholInit();
+
+	// In hatoholInit(), ConfigManager::reset() is invoked with
+	// loadConfigFile = true. This makes the test results depned on
```
cosmo0920: Nit: typo

```diff
-depned
+depend
```

cosmo0920:
I added a nitpick comment. Otherwise, it looks good to me.

kz0817:
Thank you for the comment. I fix it and send PR again.

PR #1634, opened 2015-10-21
[testFaceRestHost] Compare as JSON object.

kz0817:
The order of elements in a JSON object is arbitrary. So the simple JSON strings comparision is sometimes inappropriate and causes false-positive.

### DIFF
--- a/server/test/Helpers.cc
+++ b/server/test/Helpers.cc
@@ -197,7 +197,7 @@ extern void _assertEqualJSONString(const string &expect, const string &actual)
 		set<string> expectMembers;
 		set<string> actualMembers;
 		expectParser.getMemberNames(expectMembers);
-		expectParser.getMemberNames(actualMembers);
+		actualParser.getMemberNames(actualMembers);
 		assertEqual(expectMembers, actualMembers);
 		for (auto name : expectMembers) {
 			auto expectType = expectParser.getValueType(name);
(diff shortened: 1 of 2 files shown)


ashie:
I intentionally avoided using this assertion.
Because:

* it can't detect verbose properties. For example, please run the test without the following line:
    * https://github.com/project-hatohol/hatohol/blob/fix-broken-test/server/test/testFaceRestHost.cc#L1053
    * It still succeeds. It's not expected result.
* The message of this assertion is too hard to understand. For example, when "numberOfEvents" is replaced with "numberOfTriggers", the message is like this:
```
testFaceRestHost:
  testFaceRestHost::test_eventsWithIncidentStatusesFilter:F
===============================================================================
Failure: testFaceRestHost::test_eventsWithIncidentStatusesFilter
<expectType == actualType>
expected: <3>
  actual: <0>
./server/test/Helpers.cc:205: _assertEqualJSONString(const string&, const string&)::__lambda0(): cppcut_assert_equal(expectType, actualType, )
./server/test/testFaceRestHost.cc:1055: void testFaceRestHost::test_eventsWithIncidentStatusesFilter(): _assertEqualJSONString(expected,arg.response)
===============================================================================
```

kz0817:
Thank you for the comment.
The reasons you mentioned are defective of the current implementation of assertEqualJSONString() and I agreed with you.
However, the test case is also insufficient because it fails on my local computer due to the order of the elements.

So I will brush up assertEqualJSONString() to address the current problems.

kz0817:
I once turn down this PR and I'll send it again after the fix.

kz0817:
> It can't detect verbose properties. For example, please run the test without the following line:
> https://github.com/project-hatohol/hatohol/blob/fix-brokentest/server/test/testFaceRestHost.cc#L1053
> It still succeeds. It's not expected result.

The reason of the above behavior is a bug of assertEqualJSONString()
due to the use of wrong variable name.

PR #1635, opened 2015-10-21
[DBAgent] Add "GROUP BY" statement

cosmo0920:
Because this statement is needed by summary SQL query like this:

```
mysql> select severity, count(severity) from incidents left join events on
events.unified_id = incidents.unified_event_id where severity in
(3,4,5) group by severity;
```

And "GROUP BY" clause place is between "WHERE" and "ORDER BY" clauses.
This clause is also omittable same as "WHERE" and "ORDER BY" etc.

Strictly speaking, "HAVING" clause can put into between "GROUP BY"
clause and "ORDER BY" clauses. But now, this clause does not implement
yet.


cosmo0920:
Thanks!

PR #1643, opened 2015-10-23
[client] Update Japanese translation

ashie:


cosmo0920:
I've noticed that this the translation for connection status button is wrongly applied. WTF!

screenshot from 2015-10-23 15 26 27

Otherwise, it looks good to me.

ashie:
Oops.
gettext has "pgettext()" function to solve such issue, and Django implements it.
I'll use it.

ashie:
I fixed it.

cosmo0920:
:+1:

PR #1650, opened 2015-10-27
[EventsView][WebUI] Add labels to minimum severity filter select box

cosmo0920:
Minimum severity filter select box becomes like this:

screenshot from 2015-10-27 17 21 15

Related to #1645.


kz0817:
Are the numbers at the top really needed ?

cosmo0920:
I think that it does not matter whether the numbers at the top is removed or left as-is.
So, I decide that I'll remove it.

PR #1653, opened 2015-10-28
[WebUI][SeverityRanksView] Implement WebUI Severity Ranks Settings

cosmo0920:
Currently, only toggle important/not important functionality is available in severity ranks page.

screenshot from 2015-10-28 15 26 59


kz0817:
I think the number of "適用" buttons should be only one or nothing.

ashie:
> only one

I've implemented this approach.

save-button

kz0817:
>> only one
> I've implemented this approach.

So pretty.

PR #1656, opened 2015-10-29
[EventsView] Improve "Change treatment" menu

ashie:
* Disable "Change treatment" menu when no item is selected
* Add an empty menu item ("----------")


cosmo0920:
:+1:
Works fine in my environment.

PR #1658, opened 2015-10-29
[EventsView] Move "Latest events" button next to "Apply filter" button

ashie:
In addition, clear "End time" field when "Latest events" is
clicked.

move-latest-button


ashie:
> In addition, clear "End time" field when "Latest events" is
> clicked.

Ooops, it doesn't apply the filter.
I'll fix it.

PR #1665, opened 2015-10-30
[EventsView] Show default filter's begin time

ashie:
Set the begin time of the default filter as the placeholder of
"Begin time" entry.


cosmo0920:
:+1:

PR #1670, opened 2015-11-03
[hap2_zabbix_api] Fix a bug of calculating event_from for Zabbix API.

kz0817:
last_info in haplib is hold as a type: unicode because the input
type of event id is it. However, the code n hap2_zabbix_api excepts
an integer type. As a result, right event_form parameter is not
calculated.


kz0817:
I turn down this PR because I wrongly named the branch name.

PR #1671, opened 2015-11-03
[hap2_zabbix_api] Fix a bug of calculating event_from for Zabbix API.

kz0817:
last_info in haplib is hold as a type: unicode because the input
type of event id is it. However, the code n hap2_zabbix_api excepts
an integer type. As a result, right event_form parameter is not
calculated.

### DIFF
--- a/server/hap2/hatohol/hap2_zabbix_api.py
+++ b/server/hap2/hatohol/hap2_zabbix_api.py
@@ -111,6 +111,8 @@ class ZabbixAPIConductor:
 
         event_id_from = None
         if direction == "ASC":
+            if isinstance(last_info, unicode) or isinstance(last_info, str):
+                last_info = int(last_info)
             if isinstance(last_info, int):
                 event_id_from = last_info + 1
             event_id_till = None


kz0817:
I nominate @Mnakagawa for this PR.

Mnakagawa:
Sure:+1:

server/hap2/hatohol/hap2_zabbix_api.py
```diff
@@ -111,6 +111,8 @@ def update_events(self, last_info=None, count=None, direction="ASC",
 
         event_id_from = None
         if direction == "ASC":
+            if isinstance(last_info, unicode) or isinstance(last_info, str):
+                last_info = int(last_info)
             if isinstance(last_info, int):
```
Mnakagawa: I think, L116 become unneeded code.

Mnakagawa: Sorry, I forgot in case of event_from is null.
This code is not necessary to remove.

PR #1673, opened 2015-11-04
[EventsView] Fix a bug that statuses of some problem events become green 

ashie:
Because event["status"] is a status of a trigger, not a event.
We should use event["type"] instead of event["status"].


ashie:
I merge it myself because it's a trivial fix.

PR #1678, opened 2015-11-04
Add "types" filter for events

ashie:
Currently "statuses" filter is used for selecting events with certain states.
But "statuses" filter is inappropriate in most case because it means "Trigger statuses", not "Event statuses".
We should use "types" filter for selecting event states.


ashie:
Thanks!

PR #1679, opened 2015-11-04
[RestResourceHost] Fix a crash bug of updating an incident

ashie:
The cause is less reference count of RestResourceHost.


ashie:
I merge it myself because it's a obvious bug.

cosmo0920:
:+1: :+1: :+1:

PR #1680, opened 2015-11-04
[RestResourceSummary] Fix the number of assigned events

ashie:
Status::DONE should be treated as assigned events.
(Souldn't be treated as unassigned events.)


ashie:
I merge it myself.
It's an obvious bug.

cosmo0920:
:+1:

PR #1683, opened 2015-11-05
[HatoholAddActionDialog] Enable to select trigger status on setting i…

ashie:
…ncident sender

Because not only TRIGGER_STATUS_PROBLEM but also
TRIGGER_STATUS_OK should be confirmed by human in most sites.


cosmo0920:
If users use Redmine backend incident server, they should be able to select normal status?

ashie:
Yes.

cosmo0920:
Looks good. :+1:

ashie:
Thanks!

PR #1686, opened 2015-11-06
[Summary] Fix incorrect summary hosts statistics

cosmo0920:
* Add counting number of registered hosts
* Fix incorrect algorithm to calculate important event occurred host percentage


cosmo0920:
Thanks!

PR #1687, opened 2015-11-09
[EventsView] Change the place of an icon on "Apply filter(s)" button

ashie:
Follow the style of other buttons.


ashie:
Trivial fix.

PR #1690, opened 2015-11-09
[EventsView] Distiguish the properties "type" and "status" of events correctly in internal code

ashie:
* status: The current trigger status which is concerned with the
          event.
* type:   The event state.

In usual cases we should use the later one for events, but the
former one was often used in the previous code.

This commit fixes this issue to avoid confusion.

Please clear your config value of "event.columns" to reset the
default setting.


ashie:
Sorry, some tests are failed.
I'll fix them.

PR #1696, opened 2015-11-10
[Operationprivilege][server] Expand privileges bit width

cosmo0920:
`1` is not `unit64_t` literal but `int32` literal.
It should use `1ULL` instead. 

Related to #1694.


cosmo0920:
Thanks!

PR #1707, opened 2015-11-12
[client] Support huge OPPRVLG flags

ashie:
Since the result of "&" operator is 32bit singed integer on JavaScript, we have to treat upper bits separately.

(But it still can't treat over 52bits correctly because "Number" of JavaScript is double, not integer.)


ashie:
Sorry, it's not enough.
We should also correct HatoholUserRolesEditor.

PR #1708, opened 2015-11-12
[client] Support huge OPPRVLG flags

ashie:
Revised #1707.

Since the result of "&" operator is 32bit singed integer on JavaScript, we have to treat upper bits separately.

(But it still can't treat over 52bits correctly because "Number" of JavaScript is double, not integer.)


cosmo0920:
:+1:

ashie:
Thanks!

PR #1710, opened 2015-11-13
[hatohol-def-src-file-generator] Add uint64_t version of toString utility function

cosmo0920:
`hatohol-db-initiator` should be able to treat over 32bit integer.
Otherwise, admin privileges wrongly disappears due to overflow.


cosmo0920:
Thanks!

PR #1719, opened 2015-11-16
[hatohol-def-src-file-generator] Fix build errors on 32bit environment

ashie:


cosmo0920:
Looks good. :+1:

ashie:
Thanks!

PR #1720, opened 2015-11-16

ashie:
"SeverityRanks" should be loaded before loading events & summary to determine colors of severities.


ashie:
Sorry this code will break tests.

PR #1723, opened 2015-11-16

ashie:
"SeverityRanks" should be loaded before loading events & summary to determine colors of severities.

Revise #1720.


ashie:
Sorry, I rebase it.

PR #1724, opened 2015-11-16

ashie:
"SeverityRanks" should be loaded before loading events & summary to determine colors of severities.  Revise #1720.


cosmo0920:
:+1:

ashie:
Thanks!

PR #1731, opened 2015-11-17
[EventsView] Customize severity labels

ashie:
label-customize1

label-customize2


ashie:
In addition, set default labels of custom incident labels on CustomIncidentLabelsView
by the same way with SeverityRanksView.

ashie:
Oops, a test is failed.
I'll fix it.

PR #1735, opened 2015-11-17

kz0817:

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -1118,6 +1118,26 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
 			    const HostInfoCache &hostInfoCache,
 			    JSONRPCError &errObj)
 {
+	auto getFirstItemGroupName = [&](string &name) {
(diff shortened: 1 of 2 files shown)


server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -1118,6 +1118,26 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
 			    const HostInfoCache &hostInfoCache,
 			    JSONRPCError &errObj)
 {
+	auto getFirstItemGroupName = [&](string &name) {
+		CHECK_MANDATORY_ARRAY_EXISTENCE("itemGroupName", errObj);
+		parser.startObject("itemGroupName");
+		size_t num = parser.countElements();
+		if (num == 0) {
+			MLPL_ERR("Empty array: itemGroupName\n");
+			errObj.addError("Empty array: itemGroupName");
```
ashie: It seems that HAPI 2.0 doesn't disallow an empty array.

kz0817: Thank you for the comment. I'll take care of it.

kz0817:
I submitted the fixed version as #1736. So I close this PR.

PR #1736, opened 2015-11-18

kz0817:

****
This PR is a fixed version of #1735.

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -1118,6 +1118,35 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
 			    const HostInfoCache &hostInfoCache,
 			    JSONRPCError &errObj)
 {
+	/**
+	 * Get olny the first element from itemGroupName array.
+	 *
+	 * NOTE: This is tentative implementation. We will have to handle
+	 *       all of the given item group names.
+	 *
+	 * @param name
+	 * The first item gropu name. If the array is empty, an empty string
+	 * is assigned.
+	 * @return true if successful. Otherwise false.
+	 */
+	auto getItemGroupName = [&](string &name) {
+		CHECK_MANDATORY_ARRAY_EXISTENCE("itemGroupName", errObj);
+		parser.startObject("itemGroupName");
+		size_t num = parser.countElements();
+		if (num == 0)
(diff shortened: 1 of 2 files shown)


server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -1138,7 +1167,8 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
 		PARSE_AS_MANDATORY("brief", itemInfo.brief, errObj);
 		parseTimeStamp(parser, "lastValueTime", itemInfo.lastValueTime, errObj);
 		PARSE_AS_MANDATORY("lastValue", itemInfo.lastValue, errObj);
-		PARSE_AS_MANDATORY("itemGroupName", itemInfo.itemGroupName, errObj);
+		if (!getItemGroupName(itemInfo.itemGroupName))
+			return false;
```
cosmo0920: For now, this line's `return false` breaks previous our err checking code.
Could you add error messages in `errObj`?

kz0817: I think CHECK_MANDATORY_ARRAY_EXISTENCE("itemGroupName", errObj)
in getItemGroupName sets errObj. Is there some things we should take care of ?

cosmo0920: Ah, my understanding is wrong. Your patch is correct.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -1118,6 +1118,35 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
...
 			    JSONRPCError &errObj)
 {
+	/**
+	 * Get olny the first element from itemGroupName array.
+	 *
+	 * NOTE: This is tentative implementation. We will have to handle
+	 *       all of the given item group names.
+	 *
+	 * @param name
+	 * The first item gropu name. If the array is empty, an empty string
```
cosmo0920: [Nitpicks] typo: gropu -> group

cosmo0920:
Except for a nitpicks, it looks good to me. :wink:

server/src/HatoholArmPluginGateHAPI2.cc

kz0817:
I sent a PR with fix and rebase as #1739. So I close this PR.

PR #1739, opened 2015-11-18

kz0817:

*****
This PR is a fixed version of #1735 and #1736.


kz0817:
Thanks!

PR #1742, opened 2015-11-18
[EventsView] Simplify the filter config UI

ashie:
Stop using accordion.

filter-config-ui


ashie:
Thanks!

kz0817:
IIt looks so good.
Let's brush up lines after the check boxes to be more simple later.

PR #1746, opened 2015-11-19
[EventsView][WebUI] Set up custom incident status label

cosmo0920:
If users set up custom incident labels like this...:

screenshot from 2015-11-19 15 23 58

In events view, these values are used in tables and select box like these:

screenshot from 2015-11-19 15 25 20

screenshot from 2015-11-19 15 26 40

### NOTE

* from NONE to DONE are always shown and be able to use
* from USER01 to USER06 are only shown when label is defined

### DIFF
--- a/client/test/browser/test_events_view.js
+++ b/client/test/browser/test_events_view.js
@@ -42,6 +42,32 @@ describe('EventsView', function() {
     "apiVersion":4,
     "errorCode":0
   }
+  var dummyIncidentStatuses = {
+    "apiVersion": 4,
+    "errorCode": 0,
+    "CustomIncidentStatuses": [
+      {
+        "id": 1,
+        "code": "NONE",
+        "label": ""
+      },
+      {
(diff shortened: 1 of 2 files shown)


client/test/browser/test_events_view.js
```diff
@@ -113,14 +139,20 @@ describe('EventsView', function() {
...
+
+  function respondEvents(eventsJson) {
+    var request = this.requests[4];
+    request.respond(200, { "Content-Type": "application/json" },
                     eventsJson);
   }
 
   function respondSummary(summaryJson) {
-    var request = this.requests[4];
+    var request = this.requests[5];
```
ashie: 参考:
ちなみに順番固定でレスポンス返してるから本体側のロード順の変更で簡単にテストが
壊れてしまうので、本当はurl見てレスポンスの内容を判断すべきだと思っている。

cosmo0920:
Thanks!

PR #1760, opened 2015-11-26
Remove verbose OperationPrivilegeType

ashie:
Although several OperationPrivilegeTypes are recently added, I think some of them aren't needed.
So I remove them in this PR.

* OPPRVLG_*_CUSTOM_INCIDENT_STATUS
    * I think they can be unified with OPPRVLG_*_INCIDENT_SETTINGS
* OPPRVLG_GET_ALL_SEVERITY_RANK
    * It's not used in Hatohol server in actual.
    * It's not needed. Because all users should be allowed to get severity ranks. Otherwise the events view doesn't work correctly.


ashie:
Sorry, testFaceRestSummary::test_summary is broken.
I'll fix it.

PR #1763, opened 2015-11-26
[client] Remove broken styles

ashie:
In the previous version following styles are applied to all ul &
li elements. It breaks the layout of navbar.

  display: inline-block;
  vertical-align: middle;

The above style is introduces at 29cac9bf to align "Latest events"
button and pager buttons. But it's not needed anymore because
"Latest events" button has been moved to other place.

### DIFF
--- a/client/static/css/hatohol.css
+++ b/client/static/css/hatohol.css
@@ -40,16 +40,6 @@ div.hatohol-pager {
   text-align: center;
 }
 
-li {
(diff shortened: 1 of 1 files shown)


ashie:
Broken layout: 

broken-layout

kz0817:
I saw the problem was fixed.

ashie:
Thanks!

PR #1766, opened 2015-11-27
[EventsView] Use the default candidates for the drop down menu of inc…

ashie:
…ident statuses when the user-defined status list is empty

It's a rare case because initial user-defined statuses are inserted
by hatohol-db-initiator and they never be removed, though.


ashie:
Thanks!

PR #1768, opened 2015-11-27
[EventsView] Support toggle abbreviating event descriptions feature

cosmo0920:
Default checkbox state is unchecked:

screenshot from 2015-11-27 17 15 05

When this checked, event descriptions are abbreviated like this:

screenshot from 2015-11-27 17 17 23

### DIFF
--- a/client/static/js/events_view.js
+++ b/client/static/js/events_view.js
@@ -623,6 +623,7 @@ var EventsView = function(userProfile, options) {
       resetQuickFilter();
     });
 
+    $("#toggle-abbreviating-event-descriptions").attr("checked", false);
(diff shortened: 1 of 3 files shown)


client/static/js/events_view.js
```diff
@@ -46,6 +46,8 @@ var EventsView = function(userProfile, options) {
   self.lastQuickFilter = {};
   self.showToggleAutoRefreshButton();
   self.setupToggleAutoRefreshButtonHandler(load, self.reloadIntervalSeconds);
+  self.abbreviateDescription = false;
```
ashie: この変数必要? 直接チェックボックスの状態見てもいいような。

cosmo0920:
Thanks!

PR #1773, opened 2015-12-01
[TriggersView] Apply user defined table color

cosmo0920:
see: https://github.com/project-hatohol/hatohol/issues/1645#issuecomment-152098571

With the following severity settings:

screenshot from 2015-12-01 17 26 34

This PR makes triggers page like this:

screenshot from 2015-12-01 17 25 10


cosmo0920:
Thanks!

PR #1783, opened 2015-12-03
[server][SessionManager] Address an invalid memory access after reset().

kz0817:
Previously session timer is alive after reset(). And reset() also
recreates a SessonManager instance that is refered by the timer
handler. As a result, invalid memory access may happen.

This patch fixed the bug.
***

Note that  one of the test on TravisCI failed. However, no valgrind reported invalid memory access for all the test on my environment. So the cause of the failure doesn't seem to be related with the SessionManger.


ashie:
> Note that one of the test on TravisCI failed. However, no valgrind reported invalid memory access for all 
> the test on my environment. So the cause of the failure doesn't seem to be related with the SessionManger.

I think so too.

ashie:
I respect you!

PR #1784, opened 2015-12-03
[test][DBAgent] Fix the access to unitialized variable.

kz0817:
Without this patch, Valgrind reports the problem like

==18336== Conditional jump or move depends on uninitialised value(s)
==18336==    at 0x794E8B4: DBAgent::TableProfile::TableProfile(char const*, ColumnDef const*, unsigned long const&, DBAgent::IndexDef const*) (DBAgent.cc:93)
==18336==    by 0xB0D887C: void testDBAgent::TestDBAgent::assertUpdateArgAdd<long, int>(long const*, unsigned long const&) (testDBAgent.cc:277)
==18336==    by 0xB0CAC0D: testDBAgent::test_updateArgAddTime_t() (testDBAgent.cc:594)
==18336==    by 0x4E7E26C: ??? (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7449F: ??? (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7455D: cut_test_case_run_with_filter (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7C2F4: ??? (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7C818: ??? (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7CBAA: cut_test_suite_run_with_filter (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E7C0B0: ??? (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E6BB5B: cut_runner_run (in /usr/lib/libcutter.so.0.1.0)
==18336==    by 0x4E6B2B0: cut_run_context_start (in /usr/lib/libcutter.so.0.1.0)

PR #1785, opened 2015-12-03
[server][SelfMonitor] Add a new class to monitor Hatohol internals (＃1745)

kz0817:
The purpose of this class is to replace the self monitoring feature
which ArmUtils is currently in chage of. The implementation si ArmUtils
is a little complicated and it is hard to propagate status change among
self monitors. This new class is newly designed to addresses these problem.
****
The proposed code is identical to the one in the previous PR. It was just rebased.

### DIFF
--- a/server/common/Monitoring.h
+++ b/server/common/Monitoring.h
@@ -31,6 +31,7 @@ enum TriggerStatusType {
 	TRIGGER_STATUS_OK,
 	TRIGGER_STATUS_PROBLEM,
 	TRIGGER_STATUS_UNKNOWN,
+	NUM_TRIGGER_STATUS,
 };
 
 enum TriggerSeverityType {
@@ -104,6 +105,7 @@ typedef TriggerIdInfoMap::const_iterator       TriggerIdInfoMapConstIterator;
 typedef std::list<TriggerIdType> TriggerIdList;
 
 enum EventType {
+	EVENT_TYPE_AUTO = -2,
 	EVENT_TYPE_ALL = -1,
 	EVENT_TYPE_GOOD,
 	EVENT_TYPE_BAD,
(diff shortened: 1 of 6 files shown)


kz0817:
Hmm... Test on Travis CI failed with the strange errors.
I'll look into them.

kz0817:
The problem has not been reproduced on my local PC.

ashie:
Hmm, it seems that they are appeared from #1784 (or #1783?).

kz0817:
This may happen when the running machine is too heavy. But really ?

    Failure: testSelfMonitor::test_constructor
    currClock: 1449110986, clock: 1449110981
    <true == currClock - clock < MAX_ALLOWD_CURR_TIME_ERROR>
    expected: <1>
    actual: <0>

kz0817:
This is similar one.

    Failure: testActionManager::test_execCommandActionCrashSoon
    <false == ctx->timedOut>
    expected: <0>
    actual: <1>

kz0817:
This is strange.
The expected code 13 means KILLED by signal. The actual 14 means KILLED and DUMPED.
Typically this code is given when the core dump setting is enabled.

    Failure: testActionManager::test_execResidentActionCrashInNotifyEvent
    <failureCode == actionLog.failureCode>
    expected: <13>
    actual: <14>

kz0817:
Hmm... I could reproduce the errors on my PC.
The cause may be the load of the machine of TravisCI.

kz0817:
I ran the test with the compiler clang. In this case, the result are also successful on my local machine.

kz0817:
I think this patch itself is no problem. please go ahead wit review.

ashie:
Yes, I'm now doing it.

kz0817:
Thank you.

PR #1789, opened 2015-12-04

kz0817:
address the above problems.

(1) Internal Error (Plugin)
(4) Broker connection


kz0817:
I'll be back.

PR #1790, opened 2015-12-04

kz0817:
address the above problems.

(1) Internal Error (Plugin)
(4) Broker connection

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -28,6 +28,7 @@
 #include <libsoup/soup.h>
 #include <Reaper.h>
 #include "ArmUtils.h"
+#include "SelfMonitor.h"
 
 using namespace std;
 using namespace mlpl;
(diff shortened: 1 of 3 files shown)


kz0817:
I ran the unit test on my local computer and it succeeded.
Although the result of TravisCI failed, please review and merge it if there's no problem.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -1141,7 +1192,7 @@ static bool parseItemParams(JSONParser &parser, ItemInfoList &itemInfoList,
 		if (num > 1) {
 			MLPL_WARN("Ignore 2nd and later itemGroups. This is "
 			          "a limitation of the current version of "
-			          "Hatohol (#1721).\n");
+			          "Hatohol (#1721).");
```
ashie: It degrades #1776.

kz0817:
I sent a revised PR as #1794. So I close this issue.

PR #1792, opened 2015-12-07
[WebUI] Show an error message when Hatohol server is not working (#1581)

kz0817:
The previous implementation never show any message and doesn't
happen anything after the login button is puressed when the
Hatohol server is down.

Although it has the basic mechasim to show the message.
However, it's never been performed due to the exception at
$.parseJSON() with an empty data.

However, parsing the response data has been meaningless,
because parseLoginResultWithError() is called only when
the HTTP respose is NOT 200 (OK).

This patch simply show an error message when this method
is called back.

### DIFF
--- a/client/static/js/hatohol_connector.js
+++ b/client/static/js/hatohol_connector.js
@@ -150,24 +150,9 @@ HatoholConnector.prototype.start = function(connectParams) {
   }
 
   function parseLoginResultWithError(XMLHttpRequest, textStatus, errorThrown) {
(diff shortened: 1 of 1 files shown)


kz0817:
With this patch, the following dialog is shown.
However, we still have a problem. Either breakdown of Hatohol server or wrong password results in showing the dialog. The latter case should show that reason. I will do it soon.

kz0817:
Thank you.

PR #1030, opened 2015-02-25
[Experimental][client][server] set arbitrary http status code

cosmo0920:
This pull request contains *experimental feature* for improving testability, as follows:

* `FaceRest::ResourceHandler::replyJSONData` and `FaceRest::ResourceHandler::replyError` can accept arbitrary HTTP status code such as **401 (Unauthorized)**, **403 (Forbidden)**, ... and so on.
  * Default HTTP status is **200 (OK)**.
* Hatohol client can handle non-200 response from Hatohol server.
* Try to send response which is non-200 status code in user authentication.

Here is failing user authentication senario:

 When failing to user authenticate, Hatohol server send **401 (Unauthorized)** response to Hatohol client(WebUI). Hatohol client just passes through Hatohol server response. Nevertheless non-200 status response from Hatohol server.
 Finally, user's browser receives **401 (Unauthorized)** response from Hatohol client(WebUI).

PR #1793, opened 2015-12-07
[server] Return HTTP code 200 on login failure. (#1581)

kz0817:
The failure of login to Hatohol Server is not HTTP error.
So we should return Code 200.

### DIFF
--- a/server/src/FaceRest.cc
+++ b/server/src/FaceRest.cc
@@ -572,14 +572,14 @@ void FaceRest::handlerLogin(ResourceHandler *job)
 	gchar *user = (gchar *)g_hash_table_lookup(job->m_query, "user");
 	if (!user) {
 		MLPL_INFO("Not found: user\n");
-		job->replyError(HTERR_AUTH_FAILED, SOUP_STATUS_BAD_REQUEST);
+		job->replyError(HTERR_AUTH_FAILED);
 		return;
 	}
 	gchar *password
 	  = (gchar *)g_hash_table_lookup(job->m_query, "password");
 	if (!password) {
 		MLPL_INFO("Not found: password\n");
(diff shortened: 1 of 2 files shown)


kz0817:
With this patch, the login failure message on Web UI becomes to be shown correctly like

cosmo0920:
:+1:

ashie:
History of changing HTTP status code for login failure:

#1030 : Introduce HTTP error codes
#1194 : Bad Request (400) -> Unauthorized(401)

ashie:
Some tests in testFaceRestUser are failed because they check HTTP status code of "/login".
Could you resolve them?

ashie:
> The failure of login to Hatohol Server is not HTTP error.

That's right.
We should return 200.

FYI:

* 400 Bad Request
    * http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1
```
The request could not be understood by the server due to malformed syntax. 
```

* 401 Unauthorized
    * http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2
```
The request requires user authentication. The response MUST include a WWW-Authenticate header field (section 14.47) containing a challenge applicable to the requested resource. 
```

So both of them are unsuitable.

kz0817:
> Some tests in testFaceRestUser are failed because they check HTTP status code of "/login".
Could you resolve them?

I've not checked yet.
I'll do it.

kz0817:
I fixed the broken tests.

ashie:
I confirmed it.
Thanks!

kz0817:
Thank you.
Tonight I'll try to fix the other failed tests on the new infrastructure.

PR #1794, opened 2015-12-07

kz0817:
address the above problems.

(1) Internal Error (Plugin)
(4) Broker connection

---

This is a revised version of #1790

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -28,6 +28,7 @@
 #include <libsoup/soup.h>
 #include <Reaper.h>
 #include "ArmUtils.h"
+#include "SelfMonitor.h"
 
 using namespace std;
 using namespace mlpl;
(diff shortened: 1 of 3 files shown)


server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -157,6 +177,69 @@ struct HatoholArmPluginGateHAPI2::Impl
...
+			monitorGateInternal,
+			monitorBrokerConn,
+		};
+		for (auto &m : monitors)
+			setAllStatusComb(m);
+
+		setupMonitorXPluginInternal();
+	}
+
+	void setupMonitorXPluginInternal(void)
```
ashie: What does "X" between "Monitor" and "PluginInternal" mean?

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -157,6 +177,69 @@ struct HatoholArmPluginGateHAPI2::Impl
...
+			for (int p = 0; p < NUM_TRIGGER_STATUS; p++) {
+				for (int c = 0; c < NUM_TRIGGER_STATUS; c++) {
+					m->setEventGenerator(
+					  static_cast<TriggerStatusType>(p),
+					  static_cast<TriggerStatusType>(c),
+					  mtx[p][c]);
+				}
+			}
+		};
+
```
ashie: It's hard to understand due to some short variable name.
Could you rename them?

m -> monitor
mtx -> matrix
p -> prev or privious
c -> curr or current or new ("new" is used in SelfMonitor.h)

ashie: > mtx -> matrix

It's still too short.
triggerStatusMatrix?

server/src/HatoholArmPluginGateHAPI2.cc
It's just a typo.

server/src/HatoholArmPluginGateHAPI2.cc
kz0817: I didn't want to break lines to keep good visibility. So I used those short names.
However, some people would think like you.
I'll use proper variable names with readable code structure.

kz0817:
I fixed the broken tests.

kz0817:
I revised the PR as #1803 and close this one.

PR #1796, opened 2015-12-07
[EventsView] Fix a bug that "Abbreviate Event Descriptions" check box isn't shown by default

ashie:
In the previous implementation the check box is placed in
"change-incident-container" which isn't shown when no incident
tracking system exists. But the check box should always be shown
because it's not concerned with incidents.


ashie:
Thanks!

PR #1800, opened 2015-12-08
[server][hap2] Add missing TLS settings

ashie:
In the previous implementation TLS settings for some HAP2 plugins
aren't enabled on initial installation, they are added only on
upgrading Hatohol DB.

* server-type-hap2-ceilometer.sql
* server-type-hap2-nagios.sql
* server-type-hap2-zabbix.sql


ashie:
They were forgotten to add  in #1385.

ashie:
Thanks!

kz0817:
I ran hatoho-db-initiator after database 'hatohol' is dropped once. Then I confirmed that TLS setting fields are shown in WebUI.

PR #1801, opened 2015-12-08
[hap2_zabbix_api] Enable to work with Zabbix 2.0

### LINKED ISSUE #1799: [hap2] hap2_zabbix_api can't get events & triggers from zabbix-2.0.3
There are 2 problem:

* "state" isn't exist in trigger objects
    * https://www.zabbix.com/documentation/2.0/manual/appendix/api/trigger/definitions
* "hosts" isn't exist in event objects
    * https://www.zabbix.com/documentation/2.0/manual/appendix/api/event/definitions

ashie:
* "state" property of trigger object isn't supported by Zabbix 2.0.
  In addition, it's not suitable propery to determine trigger's
  state, we should use "value" property instead. See the following
  page for more details:

  https://www.zabbix.com/documentation/2.2/manual/api/reference/trigger/object

* In Zabbix 2.0, array of property for "selectHosts" on getting
  events isn't supported:

  https://www.zabbix.com/documentation/2.0/manual/appendix/api/event/get

  Instead require hosts on getting triggers.

Fix #1799

### DIFF
--- a/server/hap2/hatohol/test/TestZabbixApi.py
+++ b/server/hap2/hatohol/test/TestZabbixApi.py
@@ -95,14 +95,15 @@ class TestZabbixApi(unittest.TestCase):
         result = self.api.get_trigger_expand_description()
         expected = {u"result": [{u"triggerid": u"1",
                  u"description": u"test_expand_description",
-                 u"state": u"0", u"priority": u"3"}]}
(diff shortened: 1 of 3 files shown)


Mnakagawa:
All test failed, but I confirm the test success on my enviroment.
It lools so good.

ashie:
Thanks!

PR #1807, opened 2015-12-10
[travis] Use apt-get instead of pip to install MySQLdb of python

ashie:
Because pip fails to build it on the new TravisCI infrastructure
(sincde Dec. 2015).


cosmo0920:
:+1: 
But, syslog output test is still failing....

ashie:
Thanks!

PR #1810, opened 2015-12-10
[UnifiedDataStore] Fix a potential bug that an invalid pointer value …

ashie:
…may be return by getInstance()

### DIFF
--- a/server/src/UnifiedDataStore.cc
+++ b/server/src/UnifiedDataStore.cc
@@ -81,8 +81,7 @@ struct UnifiedDataStore::Impl
 		}
 	};
 
-	static UnifiedDataStore *instance;
-	static Mutex             mutex;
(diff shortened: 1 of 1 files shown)


cosmo0920:
Looks good.
But I have a question.
Should we release stored `UnifiedDataStore::Impl::instance` or set `NULL` into it?

kz0817:
NAK.

When the instance is NULL and UnifiedDataStore::getInstance() is called on two thread at the same time. The both threads do create the instance. We don't assume two instances are created.

ashie:
> When the instance is NULL and UnifiedDataStore::getInstance() is called on two threads at the same time. The both threads are possible to create instances respectively. We don't assume that two instances are created.

Sorry, I noticed it while I had dinner.
I'm fool ;-(

PR #1809, opened 2015-12-10
Error Notification of communication with AMQP Broker.

kz0817:
These patch adds a self monitor for AMQP connection.

PR #1813, opened 2015-12-10
Error Notification of communication with AMQP Broker. (2nd try)

kz0817:
These patch adds a self monitor for AMQP connection.

*******
This is a revised version of #1809.


cosmo0920:
I've confirmed that "when connecting with AMQP broker at Hatohol Server starting up, failure event does not generate" is fixed in my environment.

kz0817:
Thank you for the check.

PR #1815, opened 2015-12-11
[UnifiedDataStore] Fix a potential bug that an invalid pointer value …

ashie:
…may be return by getInstance()

Revise #1810

### DIFF
--- a/server/src/UnifiedDataStore.cc
+++ b/server/src/UnifiedDataStore.cc
@@ -82,7 +82,7 @@ struct UnifiedDataStore::Impl
 	};
 
 	static UnifiedDataStore *instance;
-	static Mutex             mutex;
+	static Mutex             getInstanceMutex;
 
 	AtomicValue<bool>        isCopyOnDemandEnabled;
 	ItemFetchWorker          itemFetchWorker;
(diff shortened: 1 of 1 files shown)


server/src/UnifiedDataStore.cc
```diff
@@ -379,14 +379,9 @@ void UnifiedDataStore::reset(void)
 
 UnifiedDataStore *UnifiedDataStore::getInstance(void)
 {
-	if (Impl::instance)
-		return Impl::instance;
-
-	Impl::mutex.lock();
+	AutoMutex lock(&Impl::getInstanceMutex);
```
cosmo0920: mlpl::AutoMutex is enough here.
But,  I think that we should use `std::lock_guard` instead.

ashie: I don't want mix different way for the same purpose in a same file.
AutoMutex is already used at another place in the same file.

We should do it in another PR.

cosmo0920: > I don't want mix different ways for the same purpose in a same file.
> AutoMutex is already used at another place in the same file.

OK. Leave it as-is.

kz0817:
I think the original code doesn't have an actual problem.
Of course, set of the pointer to the variable: instance shall be done in atomic.
Actually the set of the pointer is atomic on x86. We have to make sure it though.
(Ex. using AtomicValue)

The original code was written to be able to return instance pointer without a lock.
However, the proposed code needs the lock even after the 2nd call.

ashie:
> I think the original code doesn't have an actual problem.

Yes, it doesn't cause actual problem in the current implementation.
I close this PR.

PR #1819, opened 2015-12-11
[mlpl][test] Flush logging message before the check.

kz0817:
In many system including TravisCI, messages logged by syslog() are
buffered by the default setting. To make sure the flush of the
logged message, this patch restarts rsyslogd.


ashie:
Great!

PR #1820, opened 2015-12-11
Fix test standard hap

kz0817:
The actual object and the expected one should be compared.
However, previous code uses assertTrue(), which would be
wrongly written.

### DIFF
--- a/server/hap2/hatohol/haplib.py
+++ b/server/hap2/hatohol/haplib.py
@@ -268,6 +268,13 @@ class MonitoringServerInfo:
         }
         return handlers[type](self.extended_info)
 
+    def __str__(self):
+        s = str()
+        s += "%s %s %s " % (self.server_id, self.url, self.type)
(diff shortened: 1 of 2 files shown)


ashie:
test_normal_run still fails:

https://travis-ci.org/project-hatohol/hatohol/jobs/96244167
```
======================================================================
FAIL: test_normal_run (TestStandardHap.TestStandardHap)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/project-hatohol/hatohol/server/hap2/hatohol/test/TestStandardHap.py", line 102, in test_normal_run
    self.assertEqual(str(actual_ms), str(expect_ms))
AssertionError: 'None' != '1 http://example.com:80 0 exampleName Admin examplePass 30 10 exampleExtraInfo'
----------------------------------------------------------------------
Ran 305 tests in 2.449s
FAILED (failures=1)
```

Does a different problem still exist?

Issue #1233, opened 2015-06-10
AutoMutex implementation is replacable with C++11 <mutex> ?


cosmo0920:
This is not **issue**, but **future work**.
Hatohol has mutex utility class which is named with `AutoMutex`.
But now, C++11 provides `<mutex>` header which contains [Mutex](http://en.cppreference.com/w/cpp/thread/mutex) and [TimedMutex](http://en.cppreference.com/w/cpp/thread/timed_mutex)  classes.

This classes provide slightly more abstract functions than provided from `libpthread`.
Although, functions' signatures and return types are different from provided from `libpthread`....

kz0817:
OK. Let us it !

cosmo0920:
Note that `AutoMutex` is already replaced with C++11's `<mutex>` except for compiler ICE generating code.

PR #1822, opened 2015-12-15
[C++11][server] Replace AutoMutex with std::lock guard<mutex>

cosmo0920:
Related to #1233.

This PR does not include replacing  `HatoholArmPluginGate.cc` because this file is for HAP1.
And `DBTables.(cc|h)`'s AutoMutex are not replaced with `std::lock guard<mutex>` because it causes following internal compiler error:

```log
testDBTables.cc: In function 'void testDBTables::test_tableInitializer()':
testDBTables.cc:210:2: internal compiler error: in create_tmp_var, at gimplify.c:479
  };
  ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.8/README.Bugs> for instructions.
```

### DIFF
--- a/server/src/ActorCollector.cc
+++ b/server/src/ActorCollector.cc
@@ -23,10 +23,10 @@
 #include <sys/wait.h>
 #include <errno.h>
 #include <cstring>
+#include <mutex>
 #include "ActorCollector.h"
 #include "DBTablesAction.h"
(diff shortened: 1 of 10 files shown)


cosmo0920:
Thanks!

PR #1824, opened 2015-12-15
[OperationPrivilege] Move ALL_PRIVILEGES & NONE_PRIGILEGE into OperationPrivilege class

ashie:
Suppress the following compile time warnings:

OperationPrivilege.h:115:37: warning: 'ALL_PRIVILEGES' defined but not used [-Wunused-variable]
  const static OperationPrivilegeFlag ALL_PRIVILEGES =

In addition, remove an unused variable.

### DIFF
--- a/server/src/RestResourceUser.cc
+++ b/server/src/RestResourceUser.cc
@@ -140,17 +140,22 @@ void RestResourceUser::handlerUser(void)
 static void addUserRolesMap(
   FaceRest::ResourceHandler *job, JSONBuilder &agent)
 {
+	using namespace StringUtils;
 	UnifiedDataStore *dataStore = UnifiedDataStore::getInstance();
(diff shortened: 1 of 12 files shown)


server/src/RestResourceUser.cc
```diff
@@ -140,17 +140,22 @@ void RestResourceUser::handlerUser(void)
 static void addUserRolesMap(
   FaceRest::ResourceHandler *job, JSONBuilder &agent)
 {
+	using namespace StringUtils;
```
cosmo0920: Why add this line? In this file, StringUtils class methods are all qualified `StringUtils::`
It seems needless.

ashie: Sorry, it's mistake.

cosmo0920: I see.

PR #1826, opened 2015-12-15
[WebUI] Add the first implementation of HatoholModal.

kz0817:
We will replace HatoholDialog that uses jQuery-UI with
HatoholModal by bootstrap's Modal. This patch is a mimimum
implementation to realize a reconnect modal, which will be
merged later.

### DIFF
--- a/client/static/js/hatohol_modal.js
+++ b/client/static/js/hatohol_modal.js
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2015 Project Hatohol
+ *
+ * This file is part of Hatohol.
+ *
+ * Hatohol is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License, version 3
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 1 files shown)


client/static/js/hatohol_modal.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  var body = $('<div class="modal-body" />');
+  content.append(body);
+  body.append(params.body);
+  self.body = body;
+
+  if (params.footer != null) {
+    var footer = $('<div class="modal-footer" />');
+    content.append(footer);
```
cosmo0920: It seems to be `footer.append(params.footer);` is missing here. 
Is this intended?

kz0817: Oh. You're right. It's a mistake.
I'll fix it.

cosmo0920: Thanks for the confirming!
I'll rebase and send HatoholModal browser test patch after this PR is merged.

kz0817:
I send the fixed version as #1829 and close this issue.

PR #1829, opened 2015-12-16
[WebUI] Add the first implementation of HatoholModal.

kz0817:
We will replace HatoholDialog that uses jQuery-UI with
HatoholModal by bootstrap's Modal. This patch is a mimimum
implementation to realize a reconnect modal, which will be
merged later.

****
This is a fixed version of #1826

### DIFF
--- a/client/static/js/hatohol_modal.js
+++ b/client/static/js/hatohol_modal.js
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2015 Project Hatohol
+ *
+ * This file is part of Hatohol.
+ *
+ * Hatohol is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License, version 3
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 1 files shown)


client/static/js/hatohol_modal.js
```diff
@@ -0,0 +1,94 @@
...
+    header.append(title);
+  }
+
+  var body = $('<div class="modal-body" />');
+  content.append(body);
+  body.append(params.body);
+  self.body = body;
+
+  if (params.footer != null) {
+    var footer = $('<div class="modal-footer" />');
```
cosmo0920: Oh, `footer` is no longer used.... :confounded: 
 
Bootstrap Modal footer html structure should be like this:

```html
<div class="modal-footer">
   <some html tags>
</div>
```

cosmo0920: I think that here around code should be like this:

```js
content.append(footer);
footer.append(params.footer);
```

kz0817: Oops.
I'll fix it.

Issue #681, opened 2014-10-06
Enable to post incidents manually


ashie:
Hatohol (14.06 or later) can post an incident to Redmine automatically when the condition of the
event is mathced.
But users may want to post an incident for an unmatched event.

kz0817:
Does this issue is solved in 16.01 ?
If not, please set the Milestone properly.

ashie:
For INCIDENT_TRACKER_HATOHOL, it has been implemented at #1837.
And it's not so important for other incident trackers.

PR #1837, opened 2015-12-16
[RestResourceHost] Implement POST method for incident

ashie:
Following two parameters are required to post an incident:

* unifiedEventId
* incidentTrackerId

Parameters of the incident are generated from the specified event.

For issue #681


ashie:
最近Travis CIでよくPower buttonポチられる...

https://travis-ci.org/project-hatohol/hatohol/jobs/97161422

```
The system is going down for halt NOW!
Power button pressed 
terminate called after throwing an instance of 'HatoholException'
  what():  <DBAgentMySQL.cc:663> Failed to connect to MySQL: test_db_hatohol: (2013) Lost connection to MySQL server during query
```

PR #1838, opened 2015-12-16
[EventsView] Fix a broken message

ashie:


client/static/js/events_view.js
```diff
@@ -684,8 +684,7 @@ var EventsView = function(userProfile, options) {
         var message = parser.getMessage();
         if (!message) {
           message =
-            gettext("An unknown error is occured on changing " +
-                    "a treatment of an event with ID: ") +
+            gettext("An unknown error occured on changing a treatment of an event with ID: ") +
```
kz0817: In this usage, 'treatment' is uncountable. So 'a' just before treatment should be removed.

ashie:
I'll be back.

PR #1833, opened 2015-12-16
[WebUI] Add a new class HatoholReconnectionModal.

kz0817:
This class shows a modal on which retry countdown is performed.

PR #1839, opened 2015-12-16
[WebUI] Add a new class HatoholReconnectionModal.

kz0817:
This class shows a modal on which retry countdown is performed.

***
This is a fixed version of #1833. The difference is the following.
- Use gettext() for messages.
- Revise the title string: Connection Failure -> Connection Error


ashie:
Please add it & hatohol_modal.js to client/Makefile.am.

kz0817:
> Please add it & hatohol_modal.js to client/Makefile.am.

OK.

PR #1842, opened 2015-12-17
[WebUI] Show a modal for connection error.

kz0817:
With this patch, the modal dialog is shown when a
connection error happens. It also try to reconnect
after a while (currently 30sec. after).

This enables users to notice the error and block
the operation.

### DIFF
--- a/client/static/js/hatohol_reconnection_modal.js
+++ b/client/static/js/hatohol_reconnection_modal.js
@@ -17,9 +17,14 @@
  * <http://www.gnu.org/licenses/>.
  */
 
+var HatoholReconnectModalVars = {
+  taskList: [],
+};
+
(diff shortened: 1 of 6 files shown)


client/static/js/hatohol_reconnection_modal.js
```diff
@@ -17,9 +17,14 @@
  * <http://www.gnu.org/licenses/>.
  */
 
+var HatoholReconnectModalVars = {
+  taskList: [],
+};
+
 var HatoholReconnectModal = function(retryFunc, errorMsg) {
   var self = this;
-  var DEFAULT_RETRY_INTERVAL = 30;
```
kz0817: Oops! fix for debugging still remains.

PR #1847, opened 2015-12-18
[EventsView] Apply user-defined incident labels to filer settings candidates

cosmo0920:
Related to #1834.

If users set up following custom incident labels,

screenshot from 2015-12-18 12 01 53

Incident filter candidates are shown like this:

screenshot from 2015-12-18 12 03 49

screenshot from 2015-12-18 12 04 14

### DIFF
--- a/client/static/js/events_view.js
+++ b/client/static/js/events_view.js
@@ -153,6 +153,7 @@ var EventsView = function(userProfile, options) {
   function start() {
     $.when(loadUserConfig(), loadSeverityRank(), loadCustomIncidentStatus()).done(function() {
       load();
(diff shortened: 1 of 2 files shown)


ashie:
Hmm, it's too bad way.
self.userConfig shouldn't be recreated & replaced.
It will cause loading user settings & event filters twice.
We should only replace filterCandidates in self.userConfig.
(Add a setter function to HatoholEventsViewConfig).

cosmo0920:
I've added a setter function to HatoholEventsViewConfig and use it.

cosmo0920:
Thanks!

PR #1853, opened 2015-12-21
[WebUI] remove string 'experimental' from plugin title

teruo-oshida:
I just removed strings.
I did nothing about test but I made server/data/*.sql and checked there is no string.

$ ./autogen.sh; ./configure
$ cd server/data
$ make
$ grep -nH experimental *.sql


cosmo0920:
:+1:

PR #1858, opened 2015-12-22
[EventsView] Enable to post new incident

ashie:
Revise #1855 to fix https://github.com/project-hatohol/hatohol/pull/1855#discussion_r48215942

When Hatohol server doesn't register an incident for an event, Hatohol client can't update the status of
the incident. This patch fixes this issue by posting a new incident before changing it's status.

Currently this feature is enabled for INCIDENT_TRACKER_HATOHOL only.
When other type of incident trackers are enabled, the feature that changing incident status isn't enabled.

### DIFF
--- a/client/static/js/events_view.js
+++ b/client/static/js/events_view.js
@@ -1193,7 +1193,7 @@ var EventsView = function(userProfile, options) {
     if (!incident)
       return html + "</td>";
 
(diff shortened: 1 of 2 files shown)


cosmo0920:
:+1:

ashie:
Thanks!

PR #1862, opened 2015-12-22
[hatohol-db-initiator] Add a default incident sender

ashie:
To make it easy to use the incident feature of Hatohol, insert an incident tracker and
an incident sender action by default.


ashie:
Thanks!

PR #1871, opened 2015-12-25
Make some self monitors stateless.

kz0817:
Some monitoring items don't have the status. They detect some issues
discontinuously. The concept trigger is not suitable form them.
This patch adds a new working mode: STATELESS_MONITOR. Monitors
with the mode mainly generates events.

Thie PR also includes the patch that makes the following two self monitors STATELESS_MONITOR.
    - monitorGateInternal
    - monitorParseError
    
In addtion, some errors are saved as event. As a result, they
are shown in WebUI.

### DIFF
--- a/server/src/HatoholArmPluginGateHAPI2.cc
+++ b/server/src/HatoholArmPluginGateHAPI2.cc
@@ -52,6 +52,16 @@ struct JSONRPCError {
 	const StringList &getErrors(void) {
 		return errors;
 	}
+
+	string getConcatenatedMessage(void) const {
+		string s;
+		for (const auto &errMsg: errors) {
+			if (!s.empty())
+				s += " || ";
+			s += errMsg;
+		}
+		return s;
+	}
 };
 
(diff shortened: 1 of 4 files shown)


cosmo0920:
Looks good.
But I have a question.
`parseExchangeProfileParams(JSONParser &parser, JSONRPCError &errObj)`'s errObj may contain error information when calling it.
Should we generate stateless parse error event there, too?

kz0817:
> parseExchangeProfileParams(JSONParser &parser, JSONRPCError &errObj)'s errObj may contain error information when calling it.

Thank you for the comment.

I agree with you and I will add parseError() in the following two method.
1. HatoholArmPluginGateHAPI2::procedureHandlerExchangeProfile();
2. ExchangeProfileCallback::onGotResponse();

cosmo0920:
:+1: 
Thanks for adding missing stateless parseError triggers.
I'll merge this PR after CIs done.

PR #1882, opened 2016-01-05
[Hap2] Resolve global variable name collision.

kz0817:
'hap' is defined as a global variable and it overwrites the package
with the same name. This patch addresses it by removing the global
variable.


cosmo0920:
Pretty good! :+1: :lipstick:

PR #1885, opened 2016-01-05
[hap2] Give a logger name prefix 'hatohol.' (#1876).

kz0817:
The previous code, __name__ of hap_zabbix_api.py and the family
is replaced with  '__main__'.
The default logging setting: hap2-logging.conf has the following
line to specify the logging target.

    qualname: hatohol

As a result, their logging APIs are not loggged.


cosmo0920:
:+1:

PR #1888, opened 2016-01-05
[WebUI] Redirect the top page to ajax_dashboard. (#1761)

kz0817:
Previous implementaion returns the content of ajax_dashboard
when the result is for the top page (e.g. /hatohol).

The browser looks that the page /hatohol has the contents.
And if the login dialog is shown at the time, the JavaScript
code save the session ID in the cookie for the page.
i.e. the path is '/'.

However, the login dialog is shown at other pages such as
/hatohol/ajax_events, the path for the cookie is '/hatohol'.
Thus the two cookies are saved and the wrong one may be used.

This patch ensure that the login is always done under /hatohol.

### DIFF
--- a/client/viewer/urls.py
+++ b/client/viewer/urls.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 Project Hatohol
+# Copyright (C) 2013,2015 Project Hatohol
 #
 # This file is part of Hatohol.
 #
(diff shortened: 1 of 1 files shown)


client/viewer/urls.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 Project Hatohol
+# Copyright (C) 2013,2015 Project Hatohol
```
ashie: Happy new year!

kz0817: Oops...

kz0817:
I sent a PR of the fixed version as #1890.

PR #1889, opened 2016-01-06
[WebUI][hap2_fluentd] Add an input box for the retry interval (#1878).

kz0817:
Previous UI didn't have a field for the retry interval.
The default value (one month) was set by default.
However, this is too long. This patch provides a field for it
and users can specify any suitable value.


kz0817:
fluentd-web-ui

PR #1893, opened 2016-01-07
[client] Set the correct label for the wild card item of filters 

ashie:
Although we use "---------" as the label of the wild card item
of each filters, it's not shown since we upgrade jQuery in our
repository.

The cause is that HatoholMonitoringView.setupHostFilters() tries
to set a wrong value to the drop down menu when the wild card item
is selected.

This commit fixes the issue.


cosmo0920:
It works well in my environment. :+1:

ashie:
Thanks!

PR #1898, opened 2016-01-07
[hap2][starter] Fix hap2 starter

Mnakagawa:
Related: #1892


Mnakagawa:
I thinlk up good idea. I will re-send PR.

PR #1902, opened 2016-01-08
[hap2][zabbixapi] Use user-defined authentication exception

cosmo0920:
I'm not sure the culture of exception treating in Python, but I think that it is useful to define used-defined exception for representing AuthenticationError.

After applied this patch, Hatohol WebUI reports more concrete error type like this:

screenshot from 2016-01-08 14 53 26


cosmo0920:
@Mnakagawa How do you think about using user-defined exception class to raise authentication error exception?

Mnakagawa:
I think, it so good.
In rabbitmqconnector.py uses user define exception too.
https://github.com/project-hatohol/hatohol/blob/master/server/hap2/hatohol/rabbitmqconnector.py#L181

I will merge this PR, after check to success tests!

cosmo0920:
Thanks!

PR #1910, opened 2016-01-12
Fix config button design in event page

BlueSkyDetector:
It's for fixing #1908.

### DIFF
--- a/client/conf/locale/ja/LC_MESSAGES/django.po
+++ b/client/conf/locale/ja/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-01-07 12:44-0600\n"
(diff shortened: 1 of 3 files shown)


cosmo0920:
I've confirmed that new Web design is applied in Event Console page like this:

screenshot from 2016-01-12 11 53 21

But translation for this button's word is not applied. Is this expected?

BlueSkyDetector:
Even after executing "make .po.mo", the translation is not applied?

BlueSkyDetector:
Sorry, I needed to include *.po files.

cosmo0920:
> Even after executing "make .po.mo", the translation is not applied?

No, translation is not applied.
"Config" translation is disabled on po for this branch. Perhaps, you should run update-po in client root directory and apply translation for Config button.

cosmo0920:
:+1:

ashie:
Oops, it should be merge by the release manager (@Mnakagawa) because it's a pull request against
a "releng" branch.

cosmo0920:
Oh..., I forgot to about it....

PR #1911, opened 2016-01-12
[client] Fix a bug that the logo image isn't shown on production envi…

ashie:
…ronment

* Add the image file to the package
* Use a relative path instead of an absolute path to use it


Mnakagawa:
I confirm to show logo_hatohol.png in rpm install environment.
Thank you for your job!

PR #1920, opened 2016-01-12
[HatoholNavi] Fix a bug that wrong URLs are built for each navbar items

ashie:


Mnakagawa:
Good job!
I confirmed to fix #1919 problems.

Mnakagawa:
I will merge after success tests.

PR #1931, opened 2016-01-13
[server][data] Fix server type SQL of Livestatus

Mnakagawa:
Remove user name and password.
Livestatus plugin does not need them.


cosmo0920:
:+1:

PR #1873, opened 2015-12-29
Remove HAP1

cosmo0920:
Related to #1469.

PR #1959, opened 2016-01-19
[16.01][HAPI1] Revive HAP1-JSON plugin

cosmo0920:
I've wrongly remove in #1873. :scream: 
HAPI1-JSON has a useful plugin. We should keep to support it.


cosmo0920:
Thanks!

PR #1993, opened 2016-02-02
[TriggersView] Use user-defined severity rank labels

cosmo0920:
It would be nice to use user-defined severity rank labels in TriggersView.

By default: 

screenshot from 2016-02-02 18 33 23

Also displaying user-defined severity ranks labels:

screenshot from 2016-02-02 18 35 54


cosmo0920:
Thanks!

PR #1999, opened 2016-02-04
[hap2] Brash up haplib

Mnakagawa:
Related: #1994


kz0817:
All tests on TravisCI failed probably due to the system problem.
However, this patch worked on my local environment. So I merge this.

Mnakagawa:
Thank you:+1:

PR #2018, opened 2016-02-09
[WebUI][client] Change the sort of server types

Mnakagawa:
Related: #1915

### DIFF
--- a/client/static/js/utils.js
+++ b/client/static/js/utils.js
@@ -119,6 +119,12 @@ function getPlugin(server) {
   return hatohol["hap_" + type];
 }
 
+function addPriority(server) {
+  plugin = getPlugin(server)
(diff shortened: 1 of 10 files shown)


Mnakagawa:
Some tests ware failed. But, almost tests were successed.

client/static/js/utils.js
```diff
@@ -119,6 +119,12 @@ function getPlugin(server) {
   return hatohol["hap_" + type];
 }
 
+function addPriority(server) {
+  plugin = getPlugin(server)
+  server.hoge =  plugin.label
```
kz0817: 'hoge' ?
Probably you've forgot to name it properly.

Mnakagawa: Ohhhhhhhhh.....
It gabage code.

PR #2021, opened 2016-02-10
[hap2][SQL] Set default parameter

Mnakagawa:
Related: #1925


Mnakagawa:
Test were failed.
I will fix it.

Mnakagawa:
The failure coused by 'npm install'
I should not fix the test.

ashie:
> The failure coused by 'npm install'

Hmm, it's an issue in a node package.
It seems that it's already resolved.

PR #2024, opened 2016-02-10
[WebUI][client] Change the sort of server types

Mnakagawa:
Related: #1915

### DIFF
--- a/client/static/js/hatohol_server_edit_dialog_parameterized.js
+++ b/client/static/js/hatohol_server_edit_dialog_parameterized.js
@@ -138,15 +138,17 @@ HatoholServerEditDialogParameterized.prototype.createMainElement = function() {
 
   function replyCallback(reply, parser) {
     var type;
+    var serverTypes = reply.serverType
 
-    if (!(reply.serverType instanceof Array)) {
+    if (!(serverTypes instanceof Array)) {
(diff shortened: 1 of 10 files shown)


client/static/js/hatohol_server_edit_dialog_parameterized.js
```diff
@@ -192,6 +194,13 @@ HatoholServerEditDialogParameterized.prototype.createMainElement = function() {
     mainDiv.append('<form id="add-server-param-form" class="form-horizontal" role="form" autocomplete="off">');
     return mainDiv;
   }
+
+  function prioritySort(serverTypes) {
```
kz0817: We use the naming convention in which a method name begins with verb.
According to it, the name becomes sortByPriority().

client/static/js/utils.js
```diff
@@ -389,6 +394,23 @@ function hasFlags(flags, flagNumbers) {
   return true;
 }
 
+function sortObjectArray(obj, key, order){
```
kz0817: missing a space between "){"

client/static/js/utils.js
```diff
@@ -389,6 +394,23 @@ function hasFlags(flags, flagNumbers) {
...
+  var num_a = -1;
+  var num_b = 1;
+  if(order === 'asc'){
```
kz0817: Also missing spaces.

PR #2030, opened 2016-02-15
Allow a little difference due to platform #2009

kz0817:
As the comment in the patch, some platform generates a space
character just after a sytle property in a tag. This patch
address the problem by comparing a HTML generated by the testee
code with the expected string by match() method based on
regular expression.

In addition, the send patch clean up the position of line breaks.


cosmo0920:
Looks good. :shipit:

PR #2041, opened 2016-02-16
[Gulpfile] Add fail reporter to report lint errors

cosmo0920:
Currently, default reporter does not return non zero exit status even if there are lint errors.
This PR introduces fail reporter to report them with exit status.
Because exit status is useful for CI task.


cosmo0920:
Thanks!

PR #2050, opened 2016-02-18
[setup-by-ansible] Install aptitude first.

kz0817:
Some simple environment such as a container soon after created
is required to have 'aptitude' to progress this setup.


cosmo0920:
I'm wondering why this patch is working for lxc or container like environment.
Using aptitude instead of apt is useful for container like environment?

kz0817:
> Using aptitude instead of apt is useful for container like environment?

I don't know which is better.
However, I met the error on the execution of roles/upgrade-server and there is an error message that requires aptitude.

cosmo0920:
OK. I understood about it.

PR #2058, opened 2016-02-18
[TriggersView] Try to use bootstrap-select for live-search

cosmo0920:
Related to #2040.

I've tried to use bootstrap-select to provide live-search for selectboxes elements.
Live-search is useful for many monitoring servers, hosts, and hostgroup and so on.

Appearance will become like this:

screenshot from 2016-02-18 17 51 38

And live-search feature works like this:

screenshot from 2016-02-18 17 51 50


ashie:
とりあえずマージする。
デザインがこれで良いかどうかは @ucksts さんに監修して頂く。

cosmo0920:
Thanks!!

PR #2060, opened 2016-02-18
Remove legacy buitin Arms (#1983)

kz0817:
How HAPI2 is a standard interface for collecting data from
monitoring systems. So the regacy builtin arms are removed.

Some tests on TravsiCI failed. But, I see these sometimes before this patch is applied.
And it may happen after the feature test was modified at de2375a2231924.
So I think the following errors are not due to this patch.
 <pre>
 Register/Unregister user role test
    assertTextExists: Edited role's name "modifiedWatcher" text exists.
In /home/travis/build/project-hatohol/hatohol/client/test/feature/register_user_roles_test.js:206
  Register/Unregister user role test
    assertExists: Find an element matching: input#deleteUserRolesButton
In /home/travis/build/project-hatohol/hatohol/client/test/feature/register_users_test.js
  Register/Unregister user test
</pre>


server/test/testFaceRestServer.cc
```diff
@@ -383,14 +415,34 @@ void test_getServerType(void)
...
+	auto extractIndex = [&](const int &type, const string &uuid) {
+		for (const auto idx: remainingIndexes) {
+			const ServerTypeInfo &svti = testServerTypeInfo[idx];
+			if (svti.type == type && svti.uuid == uuid) {
+				remainingIndexes.erase(idx);
+				return idx;
+			}
+		}
+		cut_fail("Not found: type: %d, uuid: %s", type, uuid.c_str());
+		return 0ul; // This line is never invoked, just to pass build.
```
cosmo0920: Adding just comment is not good.

How about name `not_invoked_id` to this constant like this?

```cpp
constexpr auto not_invoked_idx = 0ul;
```

kz0817: Do you suggest using the constant for all tests ?
I think that using it only for this function is a little extensive.

kz0817: Or do we add a macro like ?

DECLARE_RETURN_INT_TO_PASS_BUILD
RETURN_INT_ANYTHING_ALTHOUGH_NEVER_INVOKED

cosmo0920: I've suggested using it for only this function.

But it is a nitpicks, I think.

ashie:
Sorry, it's conflicts with master.
Cloud you resolve it?

kz0817:
> Sorry, it conflicts with master.
> Could you resolve it?

OK. I'll do it soon.
So I once close this issue.

PR #2066, opened 2016-02-20
[HatoholAddActionDialog][Lint] Fix lint errors

cosmo0920:
Related to #2044.

Note that this PR contains "adding missing `break`s in `setupSevertyCompTypeValue()`".
But I think that this missing them is a bug but adding them breaks backward compatibility.
How do we treat them?


cosmo0920:
Thanks!!

ashie:
> But I think that this missing them is a bug but adding them breaks backward compatibility.
> How do we treat them?

Obbiously it's a bug.

PR #2068, opened 2016-02-20
[HatoholServerEditDialogParametrized][Lint] Use function top level var

cosmo0920:
Follows up https://github.com/project-hatohol/hatohol/pull/2054.
Also related to #2044.


cosmo0920:
:yum:

PR #2097, opened 2016-02-25
Remove legacy buitin Arms (#1983)

kz0817:
Now HAPI2 is a standard interface for collecting data from
monitoring systems. So this patch removes regacy builtin arms.

This change involves two major fixes internally.
1. Some tests using legacy arms has been changed to use HAPI2 based arms.
2. Some API such as registering a monitoring server requires an ArmPluginInfo
   instance mandatorily, which was used to be an option.


cosmo0920:
:+1:

kz0817:
Thank you.

PR #2108, opened 2016-02-26
Add a missing import.

kz0817:


cosmo0920:
It seems to be duplicated with #2106.

kz0817:
> It seems to be duplicated with #2106.

Oh! I hasn't noticed it.
Thank you for the comment.

PR #2109, opened 2016-02-26
[test][FaceRestTestUtils] Fix invalid memory reference

ashie:
Many tests fail on Ubuntu 15.10 due to this bug.


cosmo0920:
Big :+1:

PR #2111, opened 2016-02-29
Accept and stores multiple itemGroups (#1765).

kz0817:
Previosly, the HAPI2.0 interface implementation of Hatohol server
has accepted only the first item group (application in ZABBIX term)
and drops the followed ones. This patch enables multiple item groups
to be handled correctly.

To realize this feature, this patch mainly adds the followings.
- adds the tables (item_categories) newly instead of
items.item_group_name, which can have only one group name.
- 'itemGroupName' and 'applications' in the Response of FaceREST
request's /item are replaced 'itemGroupNames' and 'ItemGroups' respectively
for the easy readability.

Note: The item groups is called item categroy in Hatohol
sever code to avoid confusion with the class ItemGroup,
which is a generic container for ItemData.

### DIFF
--- a/client/static/js/latest_view.js
+++ b/client/static/js/latest_view.js
@@ -106,18 +106,18 @@ var LatestView = function(userProfile) {
 
   function parseData(replyData) {
     var parsedData = {};
-    var appNames = [];
+    var itemGrpNames = [];
     var x, item;
 
-    for (x = 0; x < replyData["applications"].length; ++x) {
-      item = replyData["applications"][x];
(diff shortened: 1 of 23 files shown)


kz0817:
Oh! Making RPM failed. I close this request and send the PR later.

cosmo0920:
It seems that building RPM failure caused by this PR: https://github.com/project-hatohol/hatohol/pull/2097.
We should adjust hatohol spec.... :cry:

PR #2114, opened 2016-03-01
[WebUI][TriggersView][BeyondMonitoringServerFilter] Add beyond monitoring server filter for hostname

cosmo0920:
This PR adds a feature which is the initial implementation beyond monitoring server filter for hostname.

`hostname` select box is newly added like this:

screenshot from 2016-03-01 16 33 07

And this select box is always selectable:

screenshot from 2016-03-01 16 33 11


cosmo0920:
Thanks!

PR #2113, opened 2016-03-01
[server] Save IncidentStatusHitory after creating/updating incidents

ashie:
To track all changes of incidents.

PR #2115, opened 2016-03-01
[server] Save IncidentStatusHitory after creating/updating incidents

ashie:
To track all changes of incidents.

Revise #2113

### DIFF
--- a/server/src/IncidentSender.cc
+++ b/server/src/IncidentSender.cc
@@ -51,7 +51,7 @@ struct IncidentSender::Job
 	    void *_userData = NULL,
 	    const UserIdType userId = USER_ID_SYSTEM)
 	: userId(userId),
-	  eventInfo(new EventInfo(_eventInfo)), incidentInfo(NULL),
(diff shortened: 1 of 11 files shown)


cosmo0920:
It would be nice to use `nullptr` where `void *userData = NULL` and `IncidentInfo *incident = NULL`.
C++11 permits writing NULL as null pointer for backward compatibility but it provides `nullptr` keyword to distinguish between NULL and null pointer.
We should use `nullptr` keyword for null pointer.

cosmo0920:
Otherwise it looks good to me! :smiley:

cosmo0920:
Can we merge this PR?

Using `nullptr` is just a suggestion. it is not mandatory for merging this PR.

ashie:
> Using nullptr is just a suggestion. it is not mandatory for merging this PR.

It should be another PR.

cosmo0920:
>> Using nullptr is just a suggestion. it is not mandatory for merging this PR.

> It should be another PR.

OK. It makes more consistency for this PR. Let's merge this PR as-is!

PR #2121, opened 2016-03-03
[TriggersView] Refresh bootstrap-select's selectpickers correctly

cosmo0920:
Refresh button has been broken since introducing `bootstrap-select` against selectboxes in triggers view.


cosmo0920:
Thanks!

PR #2140, opened 2016-03-07
[server] Rename IncidentStatusHistory to IncidentHistory

ashie:
I hate a long name.


cosmo0920:
... but I dislike short name.
Is `IncidentHistory` enough name?

ashie:
It's stores not only "status" but also "comment".
In addition other parameters of an incident such as "asignee", "priority" and "doneRatio"
should also be stored to it to remember the history.

Therefore "IncidentHistory" is more appropriate name than "IncidentStatusHistory".

ashie:
Thanks!

PR #2144, opened 2016-03-08
[tset][TriggersView] Fix an issue that a test fails on certain environments

### LINKED ISSUE #2009: [WebUI] トリガー画面のテストが環境によっては通らないことがある
以下のどれかで発生するようになったようだ。

https://github.com/project-hatohol/hatohol/pull/1995/files
https://github.com/project-hatohol/hatohol/pull/1993/files
https://github.com/project-hatohol/hatohol/pull/1992/files

原因は動的に変更している属性を含むHTMLを静的なHTMLでマッチさせようとしているため。

ashie:
The test 'With a problem trigger' fails on certain environments.

The cause is that the test expects a fixed HTML string although
it's modified by jQuery dynamically.
To resolve it, we assert only existence of "severity5" class and
backgroud color in this test. Because other parameters are tested
in other tests, we don't need test them in this test.

Fix #2009

### DIFF
--- a/client/test/browser/test_triggers_view.js
+++ b/client/test/browser/test_triggers_view.js
@@ -251,28 +251,17 @@ describe('TriggersView', function() {
 
   it('With a problem trigger', function() {
     var view = new TriggersView(getOperator(), testOptions);
-
-    // Some platofrom such as Ubuntu 14.04.2, the obtained string has a space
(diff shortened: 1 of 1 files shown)


ashie:
誰も見てくれないので @cosmo0920 さんにアサイン

ashie:
あざっす！

PR #2160, opened 2016-03-11
[hap2][hapcommon] Add try section that become effective on interractive mode

Mnakagawa:
Previously, we do not have import hap2 module on interactive mode.
If merge this patch, we will become execute it.


cosmo0920:
:+1:

ashie:
反省して面倒臭がらずにテスト結果を待つ

PR #2163, opened 2016-03-11
[configure] Remove "--enable-cpp11" switch

ashie:
Because --disable-cpp11 doesn't work anymore.
We alywas use C++11.


cosmo0920:
LGTM

ashie:
Thanks!

PR #2170, opened 2016-03-14
Webui use conf file for settings #2152 1

kz0817:
Django framework provides a way to set some parameters such as
database server and the port. However, it's a python file and
hard to be edited by users.

So this patch adds a mechanism to try to read a configuration file
under the sysconfdir given by autoconf.

### DIFF
--- a/client/conf/webui.conf
+++ b/client/conf/webui.conf
@@ -1,0 +1,12 @@
+[generic]
+#allowed_hosts = *
+#allowed_hosts = localhost 192.168.15.* *.example.com
+#TIME_ZONE = 'America/Chicago'
+
+[database]
+#name = hatohol_client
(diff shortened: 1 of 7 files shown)


client/conf/webui.conf
```diff
@@ -0,0 +1,12 @@
+[generic]
+#allowed_hosts = *
+#allowed_hosts = localhost 192.168.15.* *.example.com
+#TIME_ZONE = 'America/Chicago'
```
ashie: It should be lower case.

kz0817: Thank you for the comment !
I'll fix it soon.

kz0817:
#2172で修正版をPRしたのでクローズします。

PR #2171, opened 2016-03-14
Use %config for configuration files in RPMs.

kz0817:
Users sometimes want to save the previous settings.
With %config in RPM's spec. file, they are saved when the RPM
is removed.


cosmo0920:
Looks good. :+1:

PR #2172, opened 2016-03-14
Webui use conf file for settings #2152 2

kz0817:
*** This is a fixed version of #2170 ***

Django framework provides a way to set some parameters such as
database server and the port. However, it's a python file and
hard to be edited by users.

So this patch adds a mechanism to try to read a configuration file
under the sysconfdir given by autoconf.

PR #2174, opened 2016-03-14
Webui use conf file for settings #2152 4

kz0817:
*** This is a fixed version of #2172 ***

Django framework provides a way to set some parameters such as
database server and the port. However, it's a python file and
hard to be edited by users.

So this patch adds a mechanism to try to read a configuration file
under the sysconfdir given by autoconf.

### DIFF
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -143,6 +143,13 @@ nobase_dist_hatohol_client_DATA = \
 	viewer/severity_ranks_ajax.html \
 	viewer/tests.py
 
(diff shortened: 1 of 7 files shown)


ashie:
The client test doesn't work due to `ConfigParser.NoSectionError: No section: 'generic'`.

https://travis-ci.org/project-hatohol/hatohol/jobs/115834684

It seems that it doesn't allow a config file which doesn't have defined sections.

kz0817:
OK. I fixed the problem in the another branch. So I close this PR once.

PR #2176, opened 2016-03-15
[EventsView] Fix indent

ashie:


ashie:
I merge it myself since it's a trivial fix.

PR #2177, opened 2016-03-15
[TriggersView] Fix filter accordion appearance

cosmo0920:
It becomes more event page like design:

screenshot from 2016-03-15 17 35 05


cosmo0920:
Thanks!

PR #2178, opened 2016-03-15
[client] Fix a bug that wrong "Location" header is returned on POST m…

ashie:
…ethod


ashie:
Trivial fix.
I merge it myself.

PR #2186, opened 2016-03-18
[RestResourceIncidnet] Fix a bug that can't post comments

ashie:
Sorry, I forgot to add tests for it...

### DIFF
--- a/server/test/testFaceRestIncident.cc
+++ b/server/test/testFaceRestIncident.cc
@@ -226,6 +226,57 @@ void test_postIncidentForUnknownEvent(void)
 	assertEqualJSONString(expectedResponse, arg.response);
 }
 
+void test_putIncidentComment(void)
+{
+	loadTestDBIncidents();
+	startFaceRest();
+
+	size_t index = 2;
+	IncidentInfo expectedIncident = testIncidentInfo[index];
+	expectedIncident.status = "IN PROGRESS";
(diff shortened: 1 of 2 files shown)


server/test/testFaceRestIncident.cc
```diff
@@ -226,6 +226,57 @@ void test_postIncidentForUnknownEvent(void)
...
+			       expectedIncident.unifiedEventId));
+	RequestArg arg(path);
+	arg.userId = findUserWith(OPPRVLG_GET_ALL_SERVER);
+	arg.request = "PUT";
+	incidentInfo2StringMap(expectedIncident, arg.parameters);
+	arg.parameters["comment"] = "Assigned to @Mnagakawa";
+
+	// check the response
+	getServerResponse(arg);
+	string expectedResponse(
```
cosmo0920: Initializing string style to be mixed in same test case is not good.
Use 

```cpp
string hoge = "piyo";
```

style instead.

cosmo0920: Hmm..., both of constructor and = operator styles are valid.
Let's leave it here as-is.

cosmo0920:
LGTM
Can I merge this?

cosmo0920:
I've merged it!

ashie:
Thanks!

PR #2193, opened 2016-03-23
Configuration file for HAP2 plugins.

kz0817:
Previously plugins need many command line arguments such as paramters for conneting AMQP brokers. This patch provides a feature in which they are given by the configuration file.

NOTE: This is a revised version of #2148.

PR #2198, opened 2016-03-25
Hap2 conf 5

kz0817:
Previously plugins need many command line arguments such as paramters to connet w/ AMQP brokers. This patch provides a feature in which they are given by the configuration file.

****
This is a fixed version of #2193


kz0817:
Thanks

PR #2201, opened 2016-03-25
[client][event_view] Hide duration column

Mnakagawa:
It is temporary commit.
After release 16.04, we add new API to Hatohol server and re-display this column.

Related: #115


ashie:
Could you add a comment to describe the reason?

Mnakagawa:
In the commit?
OK

PR #2214, opened 2016-03-28
[WebUI][EventsView] Enable user comment by default

ashie:
https://github.com/project-hatohol/hatohol/issues/2119#issuecomment-202198354


ashie:
Thanks!

PR #2220, opened 2016-03-29
[EventsView] Implement links in the summary bar

ashie:
Issue #2184


ashie:
In addition, we may have to modify the UI to indicate selected filter.

cosmo0920:
It works in my environment.
Let's wait for CI results.

PR #2143, opened 2016-03-08
[server] Remove unused options "--{enable,disable}-copy-on-demand"

ashie:
It's not used in the current code.

This commit doesn't remove enable_copy_on_demand column in the DB
to avoid troublesome migration.

Fix #2142

PR #2221, opened 2016-03-30
[FaceRestIncident][test] Suppress a unused variable warning

cosmo0920:
This variable has been left since this commit: https://github.com/project-hatohol/hatohol/commit/48c950c0ea8d9fbb5162d91208cb00576d3d4901#diff-85b92f69fd18823fb7079177939e0583
Previously, cut_teardown needs to tidy up CopyOnDemand releated
settings.
But current implementation does not use CopyOnDemand related code.
It had been removed completely at #2143.


ashie:
Thanks!

PR #2224, opened 2016-03-31
Hatohol tracer

kz0817:
This patches provide a feature that calls back listers when the prefixed
trace points are invoked.

The motivation of this feature is to solve the problem #2052, in which
the intended page transition should be distinguished and allowed.


ashie:
Probably some domestic links are still remained in "Latest data" page and "Graphs" page.

kz0817:
> Probably some domestic links are still remained in "Latest data" page and "Graphs" page.

Thank you for the comment.
Let's replace them w/ domesticLink() too.

kz0817:
Thank you for the quick review.

PR #2225, opened 2016-03-31
[test][EventsView] Fix a broken test

ashie:


ashie:
The test has passed and there is no side effect.
I merge it myself.

PR #2229, opened 2016-04-03
Prevent new window from opening

kz0817:
Some links in WebUI such as menu items should not open an another a window
or a tab in order to prevent an expected situation.
This PR realizes it by forcing its own page transition method domesticLink().


kz0817:
あー、HTMLを修正したことで、大量にbrowserテストが失敗。出直します。

PR #2233, opened 2016-04-03
[WebUI] Prevent internal links from opening on a new window

kz0817:
Some links in WebUI such as menu items should not open an another a window
or a tab in order to prevent an expected situation.

This PR realizes it by forcing its own page transition method domesticLink().

NOTE: This is the fixed version of #2229


kz0817:
Thank you.

PR #2237, opened 2016-04-05
Add missing domestic links #2235

kz0817:
This pull request prevents the page transition confirmation from showing,
which has been leaked since the mechanism was taken.


client/static/js/history_view.js
```diff
@@ -60,7 +60,7 @@ var HistoryView = function(userProfile, options) {
       if (self.grpahId)
         return;
       if (!self.config.histories || self.config.histories.length <= 0)
```
Mnakagawa: ADDでcreate-modeに入って，そのまま何もせずcloseを押すと同様に確認ダイアログが発生する。
このIf文の条件いじれば治りそう

PR #2252, opened 2016-04-14
[client][server_view] Hide bulk upload button

Mnakagawa:
Currently, the function is broken. We can not fix it till 16.04 release.
It is temporary commit till 16.12 and later release.

### DIFF
--- a/client/static/js/servers_view.js
+++ b/client/static/js/servers_view.js
@@ -28,7 +28,8 @@ var ServersView = function(userProfile) {
 
   if (userProfile.hasFlag(hatohol.OPPRVLG_CREATE_SERVER)) {
     $("#add-server-button").show();
-    $("#bulkupload-server-button").show();
+//    ToDo: We fix it in 16.12 and later release.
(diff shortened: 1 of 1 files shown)


client/static/js/servers_view.js
```diff
@@ -28,7 +28,8 @@ var ServersView = function(userProfile) {
 
   if (userProfile.hasFlag(hatohol.OPPRVLG_CREATE_SERVER)) {
     $("#add-server-button").show();
-    $("#bulkupload-server-button").show();
+//    ToDo: We fix it in 16.12 and later release.
```
cosmo0920: `ToDo` should be `TODO`.

We fix it in 16.12 and later release. ->
We **should** fix this in 16.12 or later.

ですかねー。

cosmo0920: `ToDo` should be `TODO`.

We fix it in 16.12 and later release. ->
We **should** fix this in 16.12 or later.

ですかねー。

PR #2259, opened 2016-04-20
[client][event_view] Save current page when update incident status

Mnakagawa:
Related: #2226


cosmo0920:
👍 

I've confirmed this patch works fine!!

cosmo0920:
Browser tests also passed in my local environment.

Mnakagawa:
Tnank you:+1:

PR #2262, opened 2016-04-22
[Server][DBTableMonitoring] Returns the correct number of items (#2213)

kz0817:
When the multiple item groups (applications) were supported,
the problem in which the returned number of the items is smaller
than that of the request was included.

This patch addresses it by excuting two SQL statements. The first
one just collects the IDs of required items. Then the sencond
one creates ItemInfoList with the IDs.


Mnakagawa:
OK. I confirmed to fix it.

Mnakagawa:
Good job! Thank you!

PR #2284, opened 2016-05-10
[hap2] Fix a typo

cosmo0920:
```log
return ->
raise
```

`return` and `raise` are confusable appearance... :cold_sweat:


cosmo0920:
Thanks again!

PR #2285, opened 2016-05-10
[hap2][tls] Add Sender

Mnakagawa:
Related #2281


kz0817:
I confirmed that the problem is addressed.

PR #2292, opened 2016-05-11
[hap2] Convert ndoutils port number correctly

cosmo0920:
Some users want to change DB listening port from 3306 and it should
convert parsed port variable's type into `int` from `str`.

Otherwise, the following exception is raised:

```log
Plugin Error: <type 'exceptions.TypeError'>, an integer is required
```


cosmo0920:
Thanks!

cosmo0920:
Ah, I've found a similar problem in hap2-livestatus.
Should we fix it?

PR #2293, opened 2016-05-11
 [hap2][nagios-livestatus] Convert ndoutils port number correctly

cosmo0920:
`mk_livestatus` uses `socket` module internally.

It uses `socket#connect` like this: https://github.com/arthru/python-mk-livestatus/blob/master/mk_livestatus/livestatus.py#L58-L61

`socket#connect` seems to require `("ip address", integer value of port_number)` tuple: https://docs.python.org/2.7/howto/sockets.html#creating-a-socket.

We should fix it by similar approach like #2292.


cosmo0920:
:yum:

PR #2306, opened 2016-05-18
[HatoholConnector][HatoholHistory] Use timeout property in Ajax

cosmo0920:
Added `connectParam.timeout` option to set timeout value(Unit: milliseconds).

### DIFF
--- a/client/static/js/history_view.js
+++ b/client/static/js/history_view.js
@@ -270,7 +270,7 @@ var HistoryView = function(userProfile, options) {
         load();
       },
       null,
-      { pathPrefix: '' });
+      { pathPrefix: '', timeout: 30000 });
   }
 
   function enableAutoReload(onClickButton) {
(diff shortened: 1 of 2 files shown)


client/static/js/history_view.js
```diff
@@ -270,7 +270,7 @@ var HistoryView = function(userProfile, options) {
         load();
       },
       null,
-      { pathPrefix: '' });
+      { pathPrefix: '', timeout: 30000 });
```
cosmo0920: `Graph` treats a bunch of data. We should increase time out value here.

cosmo0920:
This is the one of choices to fix ~~(or conceal?)~~ waiting forever graphs rendering problem.
But this fix aims to fix **potentially problem** waiting for Ajax response forever.

cosmo0920:
Thanks!

PR #2320, opened 2016-05-24
Set armInfo status and running when timeout connection to hap2

Mnakagawa:


cosmo0920:
What is the related issue?

Mnakagawa:
This patch has no related issue.
I explain it.

In case of  hap2 connection faild, Hatohol server does not change [Connection status] to Error on [Monitoring server] screen.
This patch fix this problem.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -631,8 +636,13 @@ struct HatoholArmPluginGateHAPI2::Impl
 		virtual void onTimeout(void) override
```
cosmo0920: `FetchProcedureCallback` does not have setArmInfoStatus or similiar method like `ExchangeProfileCallback`'s `setArmInfoStatus`.
It is strange for me.
Because once `ExchangeProfileCallback` executed, it does not recover normal state except for calling  `ExchangeProfileCallback`'s `setArmInfoStatus`.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -593,9 +600,17 @@ struct HatoholArmPluginGateHAPI2::Impl
 
 		bool isSucceeded(JSONParser &parser)
 		{
+			ArmStatus &status = m_impl.m_armStatus;
+			ArmInfo armInfo = status.getArmInfo();
```
cosmo0920: I understood what you want, but I'm not agree this patch.
Because `isSucceeded` should have one functionality which is checking response JSON `result` key's value.
If you want to set armInfo status in this method, we should use another name for this, like `validateSucceeded` or similar method name.

server/src/HatoholArmPluginGateHAPI2.cc
```diff
@@ -601,6 +608,21 @@ struct HatoholArmPluginGateHAPI2::Impl
 			return result == "SUCCESS";
 		}
 
+		void updateArmInfoStatus(JSONParser &parser) {
+			ArmStatus &status = m_impl.m_armStatus;
+			ArmInfo armInfo = status.getArmInfo();
+
+			if (parser.isMember("error")) {
```
cosmo0920: Why not use `isSucceeded(parser)` here?

cosmo0920:
LGTM.

PR #2321, opened 2016-05-24
[EventsView][TriggersView] Add missing selectPicker refreshing things

cosmo0920:
This PR reduces potential issue that previous values are not to be stored in selectPickers.
But for now, I'm not sure what really causes this problem.... :/


cosmo0920:
Thanks!

PR #2333, opened 2016-05-31
[HatoholMonitoringView] Fix a wrong query name

cosmo0920:
This typo causes the problem that hostname query does not restore
correctly after filtering.
host'N'name should be hostname.
Otherwise, $("#select-hostname").val(query.hostName) sets empty string
into the specified selectbox.


cosmo0920:
Thanks!

PR #2351, opened 2016-07-04
[zabbixapi] Consider about $ macro is not digit

Mnakagawa:
Related: #2341

### DIFF
--- a/server/hap2/hatohol/zabbixapi.py
+++ b/server/hap2/hatohol/zabbixapi.py
@@ -93,7 +93,8 @@ class ZabbixAPI:
         def proc(item):
             if item["lastclock"] == "0":
                 return
-            self.expand_item_brief(item)
+            if "$" in item["name"]:
+                self.expand_item_brief(item)
 
             time = \
                 hapcommon.translate_unix_time_to_hatohol_time(item["lastclock"],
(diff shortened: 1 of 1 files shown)


cosmo0920:
Looks good.
Could you check this, @bynnchapu ?

bynnchapu:
OK. I'll check the code and behavior.

bynnchapu:
It seems code is good.
But, now I'm trying to create testing environment although, I can not add monitoring server using HAPI2...

I think it takes time to check behavior...
Please wait a little.

bynnchapu:
I've tried in the problem environment.
It seems the error log which is output in #2341 is not output but, other message is output:

```
python[20594]: hap.py:96 hatohol.hap:hap2_zabbix_api.py:Poller: [20594]: [ERROR] Unexpected error: <type 'exceptions.AssertionError'>,
                Traceback (most recent call last):
                  File "/lib/python2.7/site-packages/hatohol/haplib.py", line 1123, in __poll_in_try_block
                    self.poll()
                  File "/usr/libexec/hatohol/hap2/hatohol/hap2_zabbix_api.py", line 156, in poll
                    self.make_sure_token()
                  File "/usr/libexec/hatohol/hap2/hatohol/hap2_zabbix_api.py", line 48, in make_sure_token
                    assert ms_info is not None
                AssertionError
python[20594]: hap2_zabbix_api.py:164 hatohol.hap2_zabbix_api:hap2_zabbix_api.py:Poller: [20594]: [ERROR] Polling: aborted.
```

As a result, the message ```Plugin Error: <type 'exceptions.AssertionError'>,``` is output continue.
I think the error message which is output in #2341 is fixed, so we can merge the code for the moment.
And, we need to create new issue related above message if this Pull Request is merged.

Mnakagawa:
HAP2  gets a monitoring server info(ex. polling time, zabbix api URL) from Hatohol server when starting.
The assertion error is caused by the monitoring server info is empty.

So, the problem has no relation to this PR.

> we need to create new issue related above message

Yes. Please create it:+1:

PR #2353, opened 2016-07-04
[item][zabbix] Correspond to zabbix 3.0

Mnakagawa:
2.x uses a query as ”itemid=” but 3.0 uses "itemids[]=".
This PR resolve it to use the both in one URL.


kz0817:
Some tests failed. The expected value should also be modified.

PR #2354, opened 2016-07-04
[client][Event] Remove current trigger status

Mnakagawa:
Related: #2314


ashie:
Oops, it breaks some tests in test_events_view.js.
We should fix them.

PR #2315, opened 2016-05-23
[HistryView][WebUI] Prevent doubly requested graph drawing requests

cosmo0920:
Slightly related to #2078.

This PR fixes doubly requested graph drawing requests.

PR #2357, opened 2016-07-05
[history_view] Invoke load in loadConfig

Mnakagawa:
When we open saved graph, history_view.js does not load saved one currently.


ashie:
Probably you should revert #2315 instead of modifying this line directly.
It seems that #2315 doesn't resolve anything, it just only breaks the function.

PR #2367, opened 2016-10-27
Fix a problem that make a test fail. (#1892)

kz0817:
A test for hap (StandardHapTestee.test_normal_run)
  sometimes failed with the following error.

AssertionError: 'None' != '1 http://example.com:80 0 exampleName Admin examplePass 30 10 exampleExtraInfo'

The root cause is that test code runs
an unexpected path by receiving SIGCHILD due to the exit of
a child process created in a previous test.

### DIFF
--- a/server/hap2/hatohol/test/TestStandardHap.py
+++ b/server/hap2/hatohol/test/TestStandardHap.py
@@ -24,6 +24,29 @@ import haplib
 from hatohol import transporter
 import multiprocessing
 import json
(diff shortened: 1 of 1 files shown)


kz0817:
うーん。npmの失敗がやっぱり多いなぁ。

Mnakagawa:
Very good

cosmo0920:
:tada:

PR #2371, opened 2016-11-22
[WebUI] Enhancement security for apache.

kz0817:
This patch adds some setting for apache in order to enhance security.


ashie:
The patch itself seems O.K.
But I wonder why we don't add them to hatohol.conf.package6...

PR #2372, opened 2016-11-22
[server] Add a config file for syslog.

kz0817:
This patch provides a syslogd configuration file to
save Hatohol server log into /var/log/hatohol/hatohol-server.log.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -245,6 +245,7 @@ server/mlpl/test/Makefile
 server/mlpl/mlpl.pc
 server/common/Makefile
 server/src/Makefile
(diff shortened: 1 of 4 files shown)


ashie:
Changes in hatohol.spec.in will be included in other PR?

kz0817:
> Changes in hatohol.spec.in in #2370 will be included in other PR?

Yes.

Previous our development frequently did so. I think its approach probably would be efficient.
Honestly, I wonder which approach is better.

kz0817:
Thank you.

PR #2316, opened 2016-05-24
[client][js][event_view] Dynamic change events to click sammary button

Mnakagawa:
And fix js.po.

PR #2376, opened 2016-11-24
Work for splitting #2370 revert #2316

kz0817:
As reported in #2316, the patches break user filter.


cosmo0920:
msgfmt complains errors: https://travis-ci.org/project-hatohol/hatohol/jobs/178541729#L1677
We shall resolve these errors before merging this.

kz0817:
Thank you for your comment.

I think I'd better do 'update po' and make sure the build succeeds on the  local.
I once cancel this PR.

PR #2378, opened 2016-11-25
Don't show "Overview : Items" page

ashie:
Because it's take too long time to fetch all items if many
monitoring servers & hosts are registered. We need a breakthrough
idea to revive it.

In this commit I uncomment a configuration "ENABLED_PAGES" in
client/hatohol/base_settings.py except "ajax_overview_items" to
disable the page, but it's may not proper way to do it. So way may
change the way in the future release.


ashie:
Oops, sorry I forgot to add missing commas.

PR #2379, opened 2016-11-25
Don't show "Overview : Items" page

ashie:
Because it's take too long time to fetch all items if many
monitoring servers & hosts are registered. We need a breakthrough
idea to revive it.

In this commit I uncomment a configuration "ENABLED_PAGES" in
client/hatohol/base_settings.py except "ajax_overview_items" to
disable the page, but it's may not proper way to do it. So way may
change the way in the future release.

### DIFF
--- a/client/hatohol/base_settings.py
+++ b/client/hatohol/base_settings.py
@@ -206,30 +206,30 @@ ENABLED_PAGES = (
     #
     # monitoring views
     #
-    #"ajax_dashboard",
-    #"ajax_overview_triggers",
+    "ajax_dashboard",
+    "ajax_overview_triggers",
     #"ajax_overview_items",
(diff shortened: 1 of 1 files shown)


kz0817:
I confirmed the menu item `Overview: Items' wasn't shown.

kz0817:
The error in Travis CI seems to be due to another reason that is not related to this PR.
Probably hatohol-syslog.conf is not included in a tar ball created by `make dist'
https://travis-ci.org/project-hatohol/hatohol/jobs/178785349#L2587

make[4]: Entering directory `/home/travis/build/project-hatohol/hatohol/hatohol-16.04/server/log'
make[4]: *** No rule to make target `hatohol-syslog.conf', needed by `all-am'.  Stop.

PR #2380, opened 2016-11-27
[server][log] Fix the failure of test/dist-check.sh on TravisCI.

kz0817:
This is a leak for fed6b499a911.
So it should be merged(squahsed) in it before the branch is merged
into master.


kz0817:
I found a tiny miss.
I once close this issue.

PR #2382, opened 2016-11-27
[client][Servers] Remove Maps columns that is less useful.

kz0817:
This patch removes the Maps column in Monitoring servers page.
Because the page is for setting server's paramters and checking them,
using it as the gate to the map differents from the original
purpose of the page.

In addition, map is only provided by Zabbix. So showing it in the
common column will cause wasted spaces when Zabbix is not used.


ashie:
Memo: Still remained in hosts page.

PR #2386, opened 2016-11-27
[client][dashboard] Remove Number of items and New values per second.

kz0817:
These two values are generally difficult to obtain efficiently.
So now they aren't used informative.


cosmo0920:
> These two values are generally difficult to obtain efficiently.

Yes. I think that it is only useful for Zabbix monitoring.
HAP2 should consider various monitoring tools.

PR #2446, opened 2016-12-19
[servers] Fix a crash when the number of items is zero.

kz0817:
syncItem() assumes there is at least one elements in an input list.
Current implementation may cause crash if the list is empty.
This patch avoids the crash by checking the number of elements.


ashie:
Memo: getTargetHostId() is the root cause of the crash.

PR #2469, opened 2016-12-21
[server][Action] Add action reexecute mechanism.

kz0817:
This patch adds a feature that executes unfinished actions again
when Hatohol server starts.

### DIFF
--- a/server/src/ActionManager.cc
+++ b/server/src/ActionManager.cc
@@ -549,6 +549,56 @@ void ActionManager::checkEvents(const EventInfoList &eventList)
 	}
 }
 
+void ActionManager::reExecuteUnfinishedAction(void)
+{
+	ThreadLocalDBCache cache;
+	DBTablesAction &dbAction = cache.getAction();
+	DBTablesMonitoring &dbMonitoring = cache.getMonitoring();
+
+	ActionLogList actionLogList;
+	const vector<int> targetStatuses{ACTLOG_STAT_QUEUING,
+	                                 ACTLOG_STAT_STARTED,
+	                                 ACTLOG_STAT_RESIDENT_QUEUING,
+	                                 ACTLOG_STAT_LAUNCHING_RESIDENT};
+
+	if (!dbAction.getTargetStatusesLogs(actionLogList, targetStatuses)) {
+		MLPL_INFO("Hatohol does not have unfinished action.\n");
+		return;
+	}
+
+	ActionLogListIterator actLogIt = actionLogList.begin();
+	for (; actLogIt != actionLogList.end(); ++actLogIt) {
+		ActionLog &actionLog = *actLogIt;
+		EventInfoList eventList;
+		EventsQueryOption eventOption(USER_ID_SYSTEM);
+
+		eventOption.setTargetServerId(actionLog.serverId);
+		EventIdType eventId = actionLog.eventId;
+		list<EventIdType> eventIds{eventId};
+		eventOption.setEventIds(eventIds);
+		dbMonitoring.getEventInfoList(eventList, eventOption);
+		if (eventList.size() == 0)
+		        continue;
+		EventInfo eventInfo = *eventList.begin();
+
+		ActionDefList actionList;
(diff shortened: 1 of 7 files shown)


server/src/DBTablesAction.cc
```diff
@@ -948,6 +980,82 @@ bool DBTablesAction::getLog(ActionLog &actionLog, const string &condition)
...
+	arg.add(IDX_ACTION_LOGS_END_TIME);
+	arg.add(IDX_ACTION_LOGS_EXEC_FAILURE_CODE);
+	arg.add(IDX_ACTION_LOGS_EXIT_CODE);
+	arg.add(IDX_ACTION_LOGS_SERVER_ID);
+	arg.add(IDX_ACTION_LOGS_EVENT_ID);
+
+	getDBAgent().runTransaction(arg);
+
+	const ItemGroupList &grpList = arg.dataTable->getItemGroupList();
+	ItemGroupListConstIterator itemGrpItr = grpList.begin();
```
cosmo0920: The following comment is just my impression.
No need to take care of it.

Since C++11, `cbegin` and `cend` members has been added in iterator member.
If we start to use `auto` in type to use value type inference, it must use to create `const_iterator`s.
But, we sometimes write type name explicitly. No need to use `cbegin()` here for now.

server/src/ActionManager.cc
```diff
@@ -549,6 +549,56 @@ void ActionManager::checkEvents(const EventInfoList &eventList)
...
+	                                 ACTLOG_STAT_STARTED,
+	                                 ACTLOG_STAT_RESIDENT_QUEUING,
+	                                 ACTLOG_STAT_LAUNCHING_RESIDENT};
+
+	if (!dbAction.getTargetStatusesLogs(actionLogList, targetStatuses)) {
+		MLPL_INFO("Hatohol does not have unfinished action.\n");
+		return;
+	}
+
+	ActionLogListIterator actLogIt = actionLogList.begin();
```
cosmo0920: How about use `auto` instead of `ActionLogListIterator` here?

server/src/DBTablesAction.h
```diff
@@ -173,9 +173,15 @@ struct ActionLog {
 	int      endTime;
 	int      failureCode;
 	int      exitCode;
+	ServerIdType serverId;
+	EventIdType  eventId;
 	uint32_t nullFlags;
 };
 
+typedef std::list<ActionLog>          ActionLogList;
```
cosmo0920: How about use using type alias which had been added since C++11?

```cpp
using newtype = oldtype;
```

server/src/DBTablesMonitoring.cc
```diff
@@ -1368,6 +1375,11 @@ const std::set<EventType> &EventsQueryOption::getEventTypes(void) const
 	return m_impl->eventTypes;
 }
 
+void EventsQueryOption::setEventIds(const std::list<EventIdType> &eventIds)
```
cosmo0920: Could you remove redundant `std::` namespace prefix?
Hatohol's server side C++ source normally use `using namespace std;` on top level.

kz0817:
OK.
I will check some points including what you mentioned.

server/src/DBTablesAction.h
kz0817: Could you, please, tell me the advantage of the new style ?

server/src/DBTablesMonitoring.cc
kz0817: OK. Thank you.

kz0817:
I send a new PR #2471 with fixes using auto type and range-based for.

server/src/DBTablesAction.h
cosmo0920: In this case, both of notations causes same result and have nothing advantages.

But, using type alias can handle `template` there:

```cpp
template<class Value>
using dict = std::map<std::string, Value>;
```

And function pointer become to be more readable:

```cpp
typedef void (*fp)(int, char);
using fp = void (*)(int, char);
```

kz0817: I see. Thank you for your explanation.

PR #2471, opened 2016-12-22
[server][Action] Reexecute action mechanism.

kz0817:
This patch adds a feature that executes unfinished actions
when Hatohol server restarts.

Currently, when the Hatohol server craashes or quits forcely,
running and queued actions aren't taken care on the restart of
Hatohol server.


kz0817:
This is a revised version of #2469

kz0817:
I found some points we can use handle as `const'.
So I'll fix it and send a PR again.

PR #2472, opened 2016-12-22
[server][Action] Reexecute action mechanism.

kz0817:
This is a revised version of #2471
====
This patch adds a feature that executes unfinished actions
when Hatohol server restarts.

Currently, when the Hatohol server craashes or quits forcely,
running and queued actions aren't taken care on the restart of
Hatohol server.

### DIFF
--- a/server/src/ActionManager.cc
+++ b/server/src/ActionManager.cc
@@ -549,6 +549,52 @@ void ActionManager::checkEvents(const EventInfoList &eventList)
 	}
 }
 
+void ActionManager::reExecuteUnfinishedAction(void)
+{
+	ThreadLocalDBCache cache;
+	DBTablesAction &dbAction = cache.getAction();
+	DBTablesMonitoring &dbMonitoring = cache.getMonitoring();
+
+	ActionLogList actionLogList;
+	const vector<int> targetStatuses{ACTLOG_STAT_QUEUING,
+	                                 ACTLOG_STAT_STARTED,
+	                                 ACTLOG_STAT_RESIDENT_QUEUING,
+	                                 ACTLOG_STAT_LAUNCHING_RESIDENT};
+
+	if (!dbAction.getTargetStatusesLogs(actionLogList, targetStatuses)) {
+		MLPL_INFO("Hatohol does not have unfinished action.\n");
+		return;
+	}
+
+	for (const auto &actionLog: actionLogList) {
+		EventInfoList eventList;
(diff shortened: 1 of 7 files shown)


server/src/ActionManager.cc
```diff
@@ -549,6 +549,52 @@ void ActionManager::checkEvents(const EventInfoList &eventList)
...
+	}
+
+	for (const auto &actionLog: actionLogList) {
+		EventInfoList eventList;
+		EventsQueryOption eventOption(USER_ID_SYSTEM);
+
+		eventOption.setTargetServerId(actionLog.serverId);
+		eventOption.setEventIds({actionLog.eventId});
+		dbMonitoring.getEventInfoList(eventList, eventOption);
+		if (eventList.empty())
```
kz0817: Do we have to show warning in this condition ?

server/src/ActionManager.cc
```diff
@@ -549,6 +549,52 @@ void ActionManager::checkEvents(const EventInfoList &eventList)
...
+		        continue;
+		auto const &eventInfo = *eventList.cbegin();
+
+		ActionDefList actionList;
+		ActionsQueryOption actionOption(USER_ID_SYSTEM);
+		ActionIdList actionIdList{actionLog.actionId};
+		actionOption.setActionIdList(actionIdList);
+		dbAction.getActionList(actionList, actionOption);
+
+		if (actionList.empty())
```
kz0817: Here too.

server/src/main.cc
```diff
@@ -244,6 +245,11 @@ int mainRoutine(int argc, char *argv[])
 
 	// setup configuration database
 	ThreadLocalDBCache cache;
+
+	// Re execute unfinished action
+	ActionManager actionManager;
+	actionManager.reExecuteUnfinishedAction();
```
cosmo0920: memo: I've noticed that Hatohol server blocks this line when bunch of unfinished actions are remained.
But for now, this implementation is no problem to re-execute unfinished actions.

server/src/ActionManager.cc
masa0612: I think it is better to show logs

server/src/ActionManager.cc
masa0612: Here too

kz0817:
I added an error messages.
Please merge the PR if there's no problem.

PR #2473, opened 2016-12-22
Work for splitting #2370 rabbitmq

kz0817:
[hap2][RabbitMQConnector] Retry on the failure of publish.
    
This patch adds a feature that retries to publish again on a failure.

### DIFF
--- a/server/hap2/hatohol/rabbitmqconnector.py
+++ b/server/hap2/hatohol/rabbitmqconnector.py
@@ -21,6 +21,8 @@
 from logging import getLogger
 import os
 import pika
+from pika import exceptions
+import traceback
 import hap
 from hatohol.transporter import Transporter
 from hatohol import hapcommon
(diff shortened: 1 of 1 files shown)


server/hap2/hatohol/rabbitmqconnector.py
```diff
@@ -134,15 +147,23 @@ def __publish(self, msg):
...
+                            "exchange": "",
+                            "routing_key": self._queue_name,
+                            "body": msg,
+                            "properties": pika.BasicProperties(
+                                        content_type="application/json")
+                            }
+        try:
+            self._channel.basic_publish(**publish_argument)
+        except exceptions.ConnectionClosed:
+            self._connect()
```
cosmo0920: It seems that this retrying feature causes retrying when failing to publish forever.
Should we add upper limit of retrying count?
Or, this patch is enough for retrying feature for now?

masa0612: > It seems that this retrying feature causes retrying when failing to publish forever.

No such thing.
This sentence retry once.

cosmo0920: Hmm..., I see.

cosmo0920:
My concern is above comment.
But it is wrong according to @masa0612 .
Let's merge this PR!
:shipit:

PR #2486, opened 2016-12-27
Work for splitting #2370 fix hostqueryopt

kz0817:
This PR fixes a problem in which the some members in
HostResourceQueryOption aren't copied in the copy
constructor.


cosmo0920:
:+1:

PR #2487, opened 2016-12-27
[server][RetResourceSummary] Exclude HOLD state.

kz0817:
Unhandled import incidents should include NONE.
Currently behavior including HOLD makes users confused.


cosmo0920:
Let's wait until finishing CI.

PR #2492, opened 2016-12-31
[WebUI][feature-test] Fix failure on TravisCI.

kz0817:
This patch adds an action that opens Settings menu in the navi bar
before clicking an item inside it. Without this operations,
no click event generates on the recent TravisCI.


cosmo0920:
Any concerns? If no concern, I'll merge this PR day after tomorrow.

kz0817:
Thank you.

PR #2509, opened 2017-02-11
[Doc] Renew an overview image of Hatohol

kz0817:
The new image shows all supported Monitorings systems; zabbix,
nagios, fluentd, and ceilometer. In addition, REDMINE is also
included.


kz0817:
I turned down this PR once to fix a little.

